PR #7, opened 2015-08-14
Tao 1045 proctor specific interface

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-1045

Implements assign/unnassign test takers to a delivery


jbout:
@krampstudio Can you confirm .js ?

PR #15, opened 2015-11-05
Format tables

jsconan:
Requires https://github.com/oat-sa/tao-core/pull/617


views/js/controller/Delivery/monitoring.js
```diff
@@ -47,6 +48,42 @@ define([
...
+        var hours = Math.floor(time / 3600);
+        var parts = [];
+
+        if (hours) {
+            parts.push(hours);
+        }
+        parts.push(_leadingZero(minutes, 2));
+        parts.push(_leadingZero(seconds, 2));
+
+        return parts.join(':');
```
ssipasseuth: I would suggest truncating the seconds (second level precision is not relevant here since the displayed data is not refresh on real time).
I would also only display minutes, so 1h07 becomes 67min since minutes are more likely to be the most relevant time scale for an item duration.

views/js/tpl/item-progress.tpl
```diff
@@ -0,0 +1,18 @@
...
+{{/with}}
+{{#with item}}
+    {{#if position}}<span class="state item-progress">{{__ 'item'}} {{position}}{{#if total}}/{{total}}{{/if}}</span>{{/if}}
+    {{#with time}}
+        {{#if display}}
+    <span class="item-time">
+            {{#if elapsedStr}}<span class="elapsed" title="{{__ 'Elapsed time'}}">{{elapsedStr}}</span>{{/if}}
+        <span class="icon icon-time"></span>
+            {{#if remainingStr}}<span class="remaining" title="{{__ 'Remaining time'}}">{{remainingStr}}</span>{{/if}}
+    </span>
```
ssipasseuth: I is tru that the customer asked for  `time remaining, and time spent on current item` but now that I see both time displyed separated with the icon, I find it not very clear. What about only displaying the elapsed time and the total time, something like : 
:hourglass: 1 min / 15 min
Open to discussion...

helpers/TestCenter.php
```diff
@@ -0,0 +1,168 @@
...
+                $id = $i + 1;
+
+                $delivery   = $deliveries[array_rand($deliveries)];
+                $testTakers = getTestTakers($delivery->getId(), $deliveryService);
+                $break      = $breaks[array_rand($breaks)];
+
+                $results[] = array(
+                    'id' => $id,
+                    'delivery' => $delivery->getLabel(),
+                    'testtaker' => self::getUserName($testTakers[array_rand($testTakers)]),
```
ssipasseuth: becareful, this will break if the array $testTakers is empty...

helpers/TestCenter.php
```diff
@@ -0,0 +1,168 @@
...
+            return $cache[$deliveryId];
+        }
+
+        $status       = array('Completed', 'Terminated', 'Pending', 'Paused', 'Running');
+        $date         = array('2015-09-16 13:04', '2015-09-21 10:23', '2015-10-06 09:34', '2015-10-18 11:43', '2015-10-29 14:53');
+        $irregularity = array('', '', 'cell phone ringing', '', '', 'sickness break / restroom for 10 min', '', '');
+        $breaks       = array(0, 0, 1, 0, 0, 2, 0, 0, 3, 0, 0);
+        $results      = array();
+
+        if (!empty($deliveries)) {
```
ssipasseuth: add check && !empty($testTakers)

helpers/Delivery.php
```diff
@@ -0,0 +1,288 @@
...
+            ),
+        );
+
+        foreach ($deliveries as $delivery) {
+            $entries[] = array_merge(array(
+                'id' => $delivery->getUri(),
+                'url' => _url('monitoring', 'Delivery', null, array('delivery' => $delivery->getUri(), 'testCenter' => $testCenter->getUri())),
+                'label' => $delivery->getLabel(),
+                'text' => __('Monitor')
+            ), $mocks[array_rand($mocks)]);
```
ssipasseuth: actually, I would recommend not randomise this information but take the mocks data in order. the first delivery will *always* have the mock data 1 and the second delivery the mock number 2, etc. If there is too much deliveries, reiterate with the first mock... indeed, I am really affraid that the client will try to play a lot with it and be confused about the changing states after refresh. It is fine to randomize, but it is crucial to be consistent. Mocks are suppose to simulate a behaviour as close as the reality as possible after all.
 This comment is also applicable for delivery/index, delivery/monitoring and report/index.

PR #18, opened 2015-11-09
Tao 1707 advanced interface

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-1707


krampstudio:
It's huge, I need time to review that...

controller/Diagnostic.php
```diff
@@ -0,0 +1,69 @@
...
+
+namespace oat\taoProctoring\controller;
+
+use oat\taoProctoring\controller\Proctoring;
+use oat\taoProctoring\helpers\Breadcrumbs;
+use oat\taoProctoring\helpers\TestCenter as TestCenterHelper;
+use oat\taoProctoring\helpers\Proctoring as ProctoringHelper;
+
+/**
+ * Proctoring Delivery controllers
```
krampstudio: not sure ...

controller/Delivery.php
```diff
@@ -0,0 +1,348 @@
...
+    /**
+     * Authorise a test taker to run the delivery
+     *
+     * @throws \common_Exception
+     * @throws \oat\oatbox\service\ServiceNotFoundException
+     */
+    public function authorise() {
+
+        $deliveryId = $this->getRequestParameter('id');
+        $testTakerId = $this->getRequestParameter('tt');
```
krampstudio: Couldn't you use `delivery` and `testtaker` as request parameter ?

controller/Delivery.php
```diff
@@ -0,0 +1,348 @@
...
+            $testTakerId = array($testTakerId);
+        }
+
+        try {
+
+            $deliveryService = $this->getServiceManager()->get('taoProctoring/delivery');
+
+            $result = true;
+            foreach($testTakerId as $ttId) {
+                $result = $deliveryService->unassignTestTaker($ttId, $deliveryId) && $result;
```
krampstudio: So if it fails to unassign for only one the response is `success : false` ?

controller/Proctoring.php
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Main method to render a view for all proctoring related controller actions
+     * 
+     * @param string $cssClass
+     * @param array $data
+     * @param array $breadcrumbs
+     * @param String $template
+     */
+    protected function composeView($cssClass, $data = array(), $breadcrumbs = array(), $template = '')
```
krampstudio: If I understand well, this function let's you aggregate all your controller's data and render the main view with unspecified _data_. But it also helps you to retrieve only the breadcrumb data when you're in a jQuery ajax context but using the same endpoint ? If yes, please separate those 2 uses contexts into different functions and even endpoints (you can still have the same method that populate the 2 endpoints)

helpers/Delivery.php
```diff
@@ -0,0 +1,315 @@
...
+
+namespace oat\taoProctoring\helpers;
+
+use oat\oatbox\user\User;
+use oat\oatbox\service\ServiceManager;
+use \core_kernel_classes_Resource;
+use \common_session_SessionManager;
+use oat\taoProctoring\model\mock\WebServiceMock;
+
+class Delivery extends Proctoring
```
krampstudio: I understand this is a mock, but all the methods present in this _helper_ are your app business so it should be formalized as services then.

views/js/controller/Delivery/monitoring.js
```diff
@@ -40,28 +42,45 @@ define([
...
+            var deliveryId = $container.data('delivery');
+            var testCenterId = $container.data('testcenter');
+            var assignUrl = helpers._url('testTakers', 'Delivery', 'taoProctoring', {delivery : deliveryId, testCenter : testCenterId});
+            var removeUrl = helpers._url('remove', 'Delivery', 'taoProctoring', {delivery : deliveryId, testCenter : testCenterId});
+            var authoriseUrl = helpers._url('authorise', 'Delivery', 'taoProctoring', {delivery : deliveryId, testCenter : testCenterId});
+            var serviceUrl = helpers._url('deliveryTestTakers', 'Delivery', 'taoProctoring', {delivery : deliveryId, testCenter : testCenterId});
+
+            var bc = breadcrumbsFactory($container, crumbs);
+
+            //@TODO format the incoming data before displaying in the datatable
```
krampstudio: agree formatting the data before would be more efficient than using the datatable transform  on each colums

views/js/controller/Delivery/monitoring.js
```diff
@@ -200,18 +226,49 @@ define([
...
+                        }
+                    }, {
+                        id: 'progress',
+                        label: __('Progress'),
+                        transform: function(value, row) {
+                            var state = row && row.state;
+                            var item = state && state.item;
+                            var time = item && item.time;
+                            if (time && time.elapsed) {
+                                //if (time.total) {
```
krampstudio: do not push comments

views/js/controller/Reporting/index.js
```diff
@@ -0,0 +1,229 @@
...
+        }
+    };
+    
+    /**
+     * Create a data range picker for reporting index
+     * 
+     * @param {JQuery} $container
+     * @param {JQuery} $list
+     */
+    function dateRangePicker(date, $container, $list){
```
krampstudio: If you're going to use this more than once, then rewrap the datatable with the datepicker into a component

views/js/controller/routes.js
```diff
@@ -1,14 +1,27 @@
 define(function(){
     return {
-        'TaoProctoring' : {
+        'TestCenter' : {
```
krampstudio: not important but to notice, you can use directly an object without a function exec : 
```js
define({ 
  'TestCenter' : { }
});
```

views/js/helper/breadcrumbs.js
```diff
@@ -14,17 +14,17 @@
...
-    'use strict';
-
-    var amdLoader = document.getElementById('amd-loader');
-    var appConfig = amdLoader.getAttribute('data-config');
-    var appUrl = amdLoader.getAttribute('data-controller');
-    var appPage = amdLoader.getAttribute('data-action');
-    require([appConfig], function() {
-         require(['jquery', 'ui', appUrl + appPage + '.js'], function($, ui) {
-            ui.startEventComponents($('.content-wrap'));
+define(['ui/breadcrumbs'], function(breadcrumbs){
```
krampstudio: move this to `views/js/component/` as this is not a helper but a new component built from the breadcrumb. So it can have it's own test...

views/templates/layout.tpl
```diff
@@ -1,10 +1,58 @@
...
+        <meta charset="utf-8">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+        <title><?= Layout::getTitle() ?></title>
+        <?= tao_helpers_Scriptloader::render() ?>
+        <link rel="stylesheet" href="<?= Template::css('tao-main-style.css', 'tao') ?>" />
+        <link rel="stylesheet" href="<?= Template::css('tao-3.css', 'tao') ?>" />
+        <link rel="stylesheet" href="<?= Template::css('proctoring.css', 'taoProctoring') ?>"/>
+        <link rel="stylesheet" href="<?= Layout::getThemeStylesheet(Theme::CONTEXT_FRONTOFFICE) ?>" />
+        <link rel="shortcut icon" href="<?= Template::img('img/favicon.ico') ?>"/>
```
krampstudio: Do you have it in your extension, otherwise specify the extension

views/templates/pages/index.tpl
```diff
@@ -0,0 +1,14 @@
+<div class="container <?= get_data('cls'); ?>"<?php
+    foreach(get_data('data') as $name => $value) {
+        echo ' data-' . $name . '="' .(is_string($value) ? $value : _dh(json_encode($value))) . '"';
```
krampstudio: the formatting should be done in the controller

views/templates/pages/index.tpl
```diff
@@ -0,0 +1,14 @@
+<div class="container <?= get_data('cls'); ?>"<?php
+    foreach(get_data('data') as $name => $value) {
```
krampstudio: the template format we agreed on is : 
```html
<?php foreach(  foreach(get_data('data') as $name => $value ) : ?>
   data-<?=$name?>="<?=$value?>"
<?php endforeach ?>
```

krampstudio:
Architecture : 
I'm not a big fan of the page reload model, it might start to be a problem when things will get slower to load. 
Could you explain me how the data are given to the client ? Only though the datatable and ajax requests ? 

Server side : 
 - You can start adding some business tests, so when you'll changes from mocks to the real implementations it will be made with less pain

Client side : 
 - Think components

views/js/helper/breadcrumbs.js
ssipasseuth: ok for moving it to js/component/ however I wonder if it is really worth adding its own test since it is so small. As you can see, it just wrap the call to breadcrumbs to DRY.

views/js/controller/Reporting/index.js
ssipasseuth: It currently only addresses the specific needs for this view. Sure, if it is gonna be reused somewhere else, it is worth creating a component to it.

helpers/Delivery.php
ssipasseuth: I had the same thoughts but since this is a real temporary things, I would prefer avoid to register a few services now and to unregister them later on (prevent useless update scripts). This only serve to isolate the temporary code that are gonna be trashed anyway from the controllers. The methods and the signatures are very likely to change and the nature of the services and their location are also yet to be defined by the backend team. So in this uncertain context, I am not sure it is relevant to register services...

controller/Delivery.php
jsconan: Woops, this is old code just taken from the previous version. It will be change soon, as we will work on this aspect during this sprint

controller/Delivery.php
jsconan: Woops, this is old code just taken from the previous version. It will be change soon, as we will work on this aspect during this sprint

controller/Proctoring.php
ssipasseuth: I believe we no longer use the ajax version so I agree that we can remove it.

controller/Diagnostic.php
ssipasseuth: oops :sweat_smile:

views/templates/pages/index.tpl
jsconan: I disagree on this point as the controller must only provide pure data to the view. The controller must not pre-render HTML code. Eventually we can convert the data to JSON directly from the controller, but we loose the data structure then, this may be a concern for potential view change.

helpers/Delivery.php
krampstudio: Ok, but then add a comment explaining your intentions.

views/templates/pages/index.tpl
krampstudio: No I meant the formatting of the value itself  `(is_string($value) ? $value : _dh(json_encode($value)))` shouldn't be made in the template.

ssipasseuth:
@krampstudio we are ready for another round of review. Please tell us if the improvements made are alright. THanks !

PR #20, opened 2015-11-13
Tao 1707 delivery executions

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-1707

Refactoring on `monitoring` page:
- list of test takers moved to the `manage` action
- the page now list delivery executions
- actions have been added to authorize, pause, terminate, report
- use real data for delivery executions


helpers/Breadcrumbs.php
```diff
@@ -145,20 +145,50 @@ public static function deliveryMonitoringAll(core_kernel_classes_Resource $testC
...
      * @param array $deliveries
      * @return array
      */
-    public static function deliveryTestTaker(core_kernel_classes_Resource $testCenter, core_kernel_classes_Resource $delivery)
+    public static function deliveryManage(core_kernel_classes_Resource $testCenter, core_kernel_classes_Resource $delivery)
+    {
+        return array(
+            'id' => 'deliveryManager',
+            'url' => _url('manage', 'Delivery', null, array('testCenter' => $testCenter->getUri(), 'delivery' => $delivery->getUri())),
+            'label' => __('Manage'),
```
ssipasseuth: I would call the action "Manage Test Takers" to make it more consistent with the other action "Add Test Takers".

helpers/Breadcrumbs.php
```diff
@@ -145,20 +145,50 @@ public static function deliveryMonitoringAll(core_kernel_classes_Resource $testC
...
-        //list also other available deliveries
         return array(
             'id' => 'deliveryMonitoring',
-            'url' => _url('testTaker', 'Delivery', null, array('testCenter' => $testCenter->getUri(), 'delivery' => $delivery->getUri())),
-            'label' => __('Add Test Takers')
+            'url' => _url('testTakers', 'Delivery', null, array('testCenter' => $testCenter->getUri(), 'delivery' => $delivery->getUri())),
+            'label' => __('Add Test Takers'),
+            'entries' => array(
+                array(
+                    'id' => 'deliveryManager',
```
ssipasseuth: what about using deliveryManage() here with an extra param to exclude the entries ?

views/scss/proctoring.scss
```diff
@@ -77,7 +77,7 @@
         }
     }
     
-    .diagnostic-index, .reporting-index, .delivery-monitoring {
+    .diagnostic-index, .reporting-index, .delivery-monitoring, .delivery-manager {
```
ssipasseuth: I would have the content width of .delivery-manager to be like the one for .delivery-testtakers (=800px) to show to the user that those two screns are related and different from the monitoring one. Plus, I would add event more visual clues to .delivery-manager and  .delivery-testtakers content, e.g. an additional border, or background. It does need to be too complex but just adding small visual differences would prevent user from being confused I think.

PR #27, opened 2015-11-20
implement reporting service

ssipasseuth:
This will  temporarily disable reporting irregularities when test is finished.
The AC in the user story does not mention in which state this action should be available : https://oat-sa.atlassian.net/browse/TAO-1772
That may leave us a little time to fix this.

### DIFF
--- a/controller/Delivery.php
+++ b/controller/Delivery.php
@@ -505,13 +505,13 @@ class Delivery extends Proctoring
 
         try {
 
-            $paused = DeliveryHelper::reportExecutions($deliveryExecution, $reason);
-            $notPaused = array_diff($deliveryExecution, $paused);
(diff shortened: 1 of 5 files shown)


controller/Delivery.php
```diff
@@ -489,6 +489,38 @@ public function pauseExecutions()
...
+        $reason = $this->getRequestParameter('reason');
+
+        if (!is_array($deliveryExecution)) {
+            $deliveryExecution = array($deliveryExecution);
+        }
+
+        try {
+
+            $paused = DeliveryHelper::reportExecutions($deliveryExecution, $reason);
+            $notPaused = array_diff($deliveryExecution, $paused);
```
jsconan: Not really important, but the variables are misnamed here...

PR #33, opened 2015-11-24
Use actual test centers instead of mock data

jsconan:


model/implementation/DeliveryService.php
```diff
@@ -90,7 +92,32 @@ public function getProctorableDeliveries(User $proctor, $options = array())
...
+    public function getTestCenterDeliveries($testCenterId)
+    {
+        if (is_object($testCenterId) && $testCenterId instanceof \core_kernel_classes_Resource) {
+            $testCenter = $testCenterId;
+            $testCenterId = $testCenter->getUri();
+        }
+
+        if (is_string($testCenterId)) {
+            $testCenter = new \core_kernel_classes_Resource($testCenterId);
+        }
```
ssipasseuth: strange construction for the resource $testCenter : what about only contructing the Resource if the testCenterId is a string ? If it is already a Resource, you can simply assign it to the variable $testCenter

jsconan: Yes, it was a bad construction. I fixed it.

PR #34, opened 2015-11-24
Add authorized_by property

hutnikau:


jsconan:
Joel is working on a better implementation for delivery sessions and storage. So I cannot validate this PR now as ExtendedStateService will probably disappear and will be replaced by something else.

hutnikau:
But i didn't use `ExtendedStateService` there. I guess that `setProctoringState()` method will not be removed from `DeliveryService`.
@jbout could you please look at this PR?

jbout:
As I am currently in the refactoring of the delivery services it is a bit too early for me to say how exactly this will impact proctoring

PR #37, opened 2015-12-04
Tao 1832 test site reporting of assessment results

hutnikau:


views/templates/Reporting/print_rubric.tpl
```diff
@@ -0,0 +1,54 @@
...
+    </head>
+    <body class="proctoring-scope">
+        <div class="assessment_results">
+            <?php foreach($rubrics as $rubricNum => $rubric): ?>
+                <?php if ($rubricNum > 0): ?>
+                    <div class="end-page"></div>
+                <?php endif; ?>
+                <header>
+                    <h1>
+                        <?= $rubric['testData']['Label'] ?>
```
jsconan: The `Label` index can be undefined, and PHP displays a notice

views/js/controller/Reporting/index.js
```diff
@@ -114,6 +131,53 @@ define([
...
+                    label: __('Download CSV'),
+                    action: function() {
+                        notYet();
+                    },
+                }, {
+                    id : 'printRubric',
+                    title : __('Print rubric block with item tagged with tao-print tag'),
+                    icon : 'print',
+                    label : __('Print Score Report'),
+                    action : printRubric
```
jsconan: This tool button should only appear when lines are selected.
Please use the property `massAction: true`.

views/js/controller/Reporting/index.js
```diff
@@ -104,7 +105,23 @@ define([
...
                 });
             };
-            
+
+            /**
+             * Open new tab with page to be printed
+             * @param string|array rowId
+             */
+            var printResults = function printReport(rowId) {
+                window.open(helpers._url('printReport',  'Reporting', 'taoProctoring', {'id' : rowId}));
```
jsconan: Please name the popup to avoid many windows opened at the same time.

```javascript
window.open(url, 'printResults');
```

views/js/controller/Reporting/index.js
```diff
@@ -104,7 +105,23 @@ define([
...
+            }
+
+            /**
+             * Print rubric blocks with item tagged with tao-print tag
+             * @param string|array rowId
+             */
+            var printRubric = function printReport(rowId) {
+                window.open(helpers._url('printRubric',  'Reporting', 'taoProctoring', {'id' : rowId}));
```
jsconan: Please name the popup

views/templates/Reporting/print_report.tpl
```diff
@@ -0,0 +1,112 @@
...
+                                <?php endif; ?>
+                            </table>
+                        </div>
+                    <?php endforeach; ?>
+                </div>
+                <div class="end-page"></div>
+            <?php endforeach; ?>
+        </div>
+        <script>
+            window.print();
```
jsconan: also make sure that this window will be closed after print

views/templates/Reporting/print_rubric.tpl
```diff
@@ -0,0 +1,54 @@
...
+                    <div class="clearfix rubric-container">
+                        <?php foreach($rubric['rubricContent'] as $rubricContent): ?>
+                            <?= $rubricContent ?>
+                        <?php endforeach; ?>
+                    </div>
+                </div>
+            <?php endforeach; ?>
+        </div>
+        <script>
+            window.print();
```
jsconan: also make sure that this window will be closed after print

model/AssessmentResultsService.php
```diff
@@ -0,0 +1,217 @@
...
+        $$basePathVarName = $compilationDirs['public']->getPublicAccessUrl();
+
+        // state name (the variable to access to get the state of the assessmentTestSession).
+        $stateName = TAOQTITEST_RENDERING_STATE_NAME;
+        $$stateName = $session;
+
+        // views name (the variable to be accessed for the visibility of rubric blocks).
+        $viewsName = TAOQTITEST_VIEWS_NAME;
+        $$viewsName = array(View::CANDIDATE);
+
```
jsconan: This looks a little weird.
Isn't it a better way to do that ?

model/AssessmentResultsService.php
```diff
@@ -0,0 +1,217 @@
...
+        foreach ($sections as $section) {
+            $assessmentItemsRef = $section->getComponentsByClassName('assessmentItemRef');
+            foreach ($assessmentItemsRef as $item) {
+                foreach($item->getCategories() as $category) {
+                    if ($category === $tag) {
+                        foreach ($section->getRubricBlockRefs() as $rubric) {
+                            ob_start();
+                            include($compilationDirs['private']->getPath() . $rubric->getHref());
+                            $rubrics[] = ob_get_clean();
+                        }
```
jsconan: I understand the needs, but is there not a better way to render these templates ?
@bugalot can you confirm ?

views/templates/Reporting/print_report.tpl
```diff
@@ -0,0 +1,112 @@
...
+                <hr>
+                <div>
+                    <p class="title">
+                        <?= __('Test Variables') ?>
+                    </p>
+                    <div class="avoid-page-break table-container clearfix">
+                        <table>
+                            <tr>
+                                <td><?= __('Label') ?></td>
+                                <td><?= $report['testData']['Label'] ?></td>
```
jsconan: Be careful the `Label` index can be undefined

views/templates/Reporting/print_report.tpl
```diff
@@ -0,0 +1,112 @@
...
+        <link rel="stylesheet" href="<?= Template::css('printReport.css', 'taoProctoring') ?>"/>
+        <link rel="stylesheet" href="<?= Layout::getThemeStylesheet(Theme::CONTEXT_FRONTOFFICE) ?>" />
+        <link rel="shortcut icon" href="<?= Template::img('favicon.ico', 'tao') ?>"/>
+    </head>
+    <body class="proctoring-scope">
+        <div class="assessment_results">
+            <?php foreach($reports as $report): ?>
+                <header>
+                    <h1>
+                        <?= $report['testData']['Label'] ?> <?=__('results')?>
```
jsconan: Be careful the `Label` index can be undefined

model/AssessmentResultsService.php
hutnikau: Actually I'm almost completely copied these lines from here:
https://github.com/oat-sa/extension-tao-testqti/blob/develop/helpers/class.TestRunnerUtils.php#L517-L539

I don't think that we can render it without using output buffering because these files (rubric templates) located in the private directory with compiled test.

hutnikau:
Branch has been updated. @jsconan please review it again.

model/AssessmentResultsService.php
bugalot: I would say this could go in a helper to make it cleaner as it is close to be the same code than in the QTI Test Driver.

However, about the technique, I don't see any other technical sugar enabling such a thing. If we look at usual template engines (Twig, Smarty, our own Template Renderer), we go for output buffering, there's no other clean way, AFAIK.

@jsconan Due to the relative urgence to have this feature working, I would ask you to merge this PR. @hutnikau, I would like you to think about putting this rendering code into an helper that proctoring and qti test can share in nice way.

PR #40, opened 2015-12-09
Tao 1831 activity report

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-1831

Requires: https://github.com/oat-sa/tao-core/pull/664

- paginate before detailing data
- moved user helpers to tao/core
- format date server side according to user's locale
- refactoring and clean up


antoinerobin:
in controller/Reporting.php can you please use the assessmentResultsService CONFIG_ID in the service manager

PR #48, opened 2015-12-15
refactored to be compatible with API

boajer:
Related to https://github.com/oat-sa/tao-core/pull/670


ssipasseuth:
You need to refactor all the current usage of ui/bulkActionPopup, including https://github.com/oat-sa/extension-tao-proctoring/blob/feature/configurable-categories-selector/views/js/controller/Delivery/monitoring.js

PR #49, opened 2015-12-16
Ta0 1942 provide service

antoinerobin:


model/ProctorManagementService.php
```diff
@@ -0,0 +1,162 @@
...
+        $proctorClass = $this->getRootClass();
+        $users = array();
+        foreach($testCenters as $testCenterUri){
+            $assignedProctors = $proctorClass->searchInstances(array(
+                self::PROPERTY_ASSIGNED_PROCTOR_URI => $testCenterUri
+            ), array(
+                'recursive' => true,
+                'like' => false
+            ));
+            $users = array_merge($users , $assignedProctors);
```
jsconan: What happens if the same proctor is assigned to several test centers ? I guess he will appear several time in the result, am I wrong ?

antoinerobin: as you can see here https://github.com/oat-sa/extension-tao-proctoring/pull/49/files#diff-d4d3195d505a239bb3cb856fd0a23e21R205 the array_merge should solve take care of it and return the proctor only once

PR #52, opened 2015-12-17
Tao 1937 proctor form

ssipasseuth:
note 1 : there is currently no unit test for the form component because it rely on server form generation + validation.
note 2 :  the form submission can create a proctor but it will fails if the test site admin is not assigned to any test center. This situation should not happen since the ui should prevent this eventually.


views/js/component/proctorForm.js
```diff
@@ -24,54 +24,117 @@ define([
...
+        users.checkLogin(formData.loginId, proctorLoginCheckUrl);
     }
-    
+
+    /**
+     * Create the form in the container
+     * @param {type} $container
+     * @returns {undefined}
+     */
     function init($container){
```
jsconan: I am a little bit confused... Is this function still in use ? The factory below makes the same thing...

views/js/component/proctorForm.js
```diff
@@ -24,54 +24,117 @@ define([
...
+                            testCenters : [] //@todo get the testCenter from the test center datalist component (config.testCenterList)
+                        };
+                        _.each(fields, function(field){
+                            data[field.name] = field.value;
+                        });
+
+                        $.post(proctorFormUrl, data, function(res){
+                            if(res.created){
+                                feedback().success(__('Proctor created'));
+                                $element.empty();
```
jsconan: I think there is something missing: we need to known when the component has done its stuff, in order to go back to the list. Maybe an event or a callback ?

controller/ProctorManager.php
```diff
@@ -180,23 +181,10 @@ public function createProctorForm(){
         
         return $this->returnJson(array(
```
jsconan: The `returnJson` method does not return a value, so it is useless to apply a `return` here.

views/js/component/proctorForm.js
ssipasseuth: Yes, let's figure it out during the integration phase.

ssipasseuth:
Thabks for your feedback, Ready for a new review !

PR #53, opened 2015-12-17
Refactor/delivery rdf

jbout:


bugalot:
Tested with:

* TAO Verbatim
* TAO Proctoring

PR #59, opened 2015-12-18
add a tooltip with authorized testcenter for partially authorized pro…

antoinerobin:
…ctor


views/js/templates/ProctorManager/status.tpl
```diff
@@ -1,2 +1,2 @@
 {{#if authorized}}{{__ 'Authorized'}} <span class="icon icon-authorize"></span>{{/if}}
-{{#if partially}}{{__ 'Partially authorized'}}{{/if}}
\ No newline at end of file
+<span title="{{testCenters}}">{{#if partially}}{{__ 'Partially authorized'}}{{/if}}</span>
```
jsconan: It could be better if you put the span inside the if, in order to avoid useless span displayed

PR #64, opened 2015-12-21
Tao 1984 delivery monitoring interface

jbout:


jbout:
Used quick and dirty alternative to represent the state description

PR #69, opened 2015-12-27
Use new delivery execution interface

jbout:


model/implementation/DeliveryService.php
```diff
@@ -26,7 +26,7 @@
 use oat\taoProctoring\model\ProctorAssignment;
 use core_kernel_users_GenerisUser;
 use oat\taoGroups\models\GroupsService;
-use oat\taoDelivery\models\classes\execution\DeliveryExecution;
+use oat\taoDelivery\model\execution\DeliveryExecution;
 use oat\taoDelivery\model\execution\DeliveryExecution as DeliveryExecutionInt;
```
hutnikau: I guess that `DeliveryExecutionInt` alias can be removed (there only four usages of this alias in this file).

PR #72, opened 2016-01-04
fixed missing path in build bundle.js file

ssipasseuth:


ssipasseuth:
wrong destination branch

PR #76, opened 2016-01-08
Tao 2011 test center context

antoinerobin:


model/implementation/DeliveryService.php
```diff
@@ -610,23 +647,36 @@ public function resumeSession(AssessmentTestSession $session)
...
         ));
         if (empty($groups)) {
-            \common_Logger::w('No system group exists for delivery '.$deliveryId.'. creating one');
-            $delivery = new \core_kernel_classes_Resource($deliveryId);
-            $newGroup = GroupsService::singleton()->getRootClass()->createInstance('test takers for delivery '.$delivery->getLabel());
-            $newGroup->setPropertyValue(new \core_kernel_classes_Property(PROPERTY_GROUP_DELVIERY), $deliveryId);
+            \common_Logger::w('No system group exists for delivery '.$deliveryId.' and test center '.$testCenterId.'. creating one');
+            $delivery = new Resource($deliveryId);
+            $testCenter = new Resource($testCenterId);
+            $instanceName = 'test takers for delivery '.$delivery->getLabel().' and Test center '.$testCenter->getLabel();
```
jsconan: Not very important, but why is `Test center` in upper case while `test takers` and `delivery` are not ?

PR #83, opened 2016-01-29
Fix/tao 2058 sub testcenters

antoinerobin:


ssipasseuth:
It works fine and the code is fine ! Please just increase the version number as Lionel suggested.

ssipasseuth:
@jsonan, please just check the version number increase and wait for the next merge window (when the develop branch is merged into master)

PR #84, opened 2016-01-29
No error feedback shown,  if list of eligible testakers was not modif…

boajer:
…ied before saving

### DIFF
--- a/views/js/component/eligibilityEditor.js
+++ b/views/js/component/eligibilityEditor.js
@@ -97,8 +97,6 @@ define([
             }
         });
 
-        instance.eligibility.deliveries = _.uniq(tree.getChecked());
-
         return tree;
     }
 
--- a/views/js/component/eligibilityEditor.js
+++ b/views/js/component/eligibilityEditor.js
@@ -39,7 +39,7 @@ define([
     /**
      * Builds group tree inside target container
      * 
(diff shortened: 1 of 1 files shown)


ssipasseuth:
Hello @boajer , sorry for the delay on this PR review.
I would need some clarification. Could you define step by step the problem you are trying to solve please ? (I am sure I understand...) Thanks !

boajer:
@ssipasseuth This is very minor issue. If you are as backend admin will open 'Edit Eligibility' popup on testcenter page and then click 'Ok' button without changing any testtakers, you will observe feedback message 'Unknown Error' on the screen, because normal output will be broken by following error 'Warning: strpos() expects parameter 1 to be string, array given in /www/act/tao/helpers/class.Uri.php on line 255'

ssipasseuth:
Sorry, it has been left aside for a while... do you stilll have this issue ? I don't seem to be able to reproduce it anymore...

boajer:
Still reproducible, i've captured small video for you http://goo.gl/K1n91K

ssipasseuth:
Thanks for the video. Strange though I didn't have that issue. But this did fix the issue when saving the eligibility editor with no test taker. It used to save an empty string. Now simply ignore the changes.

PR #90, opened 2016-02-12
Added delivery execution terminated event

jbout:


hutnikau:
Looks good for me but i think it should be merged after listener of this event will be implemented to save proper exit code.

PR #91, opened 2016-02-19
Fix the wrong use of TestSessionService in the reporting helpers

jsconan:
The reporting was broken in the proctoring interface.

### DIFF
--- a/model/AssessmentResultsService.php
+++ b/model/AssessmentResultsService.php
@@ -42,24 +42,6 @@ class AssessmentResultsService extends ConfigurableService
     const OPTION_PRINT_REPORT_BUTTON = 'print_report_button';
 
     /**
-     * @var TestSessionService
-     */
-    private $testSessionService;
-
-    /**
-     * Gets test session service
-     *
-     * @return TestSessionService
-     */
-    private function getTestSessionService()
-    {
-        if ($this->testSessionService === null) {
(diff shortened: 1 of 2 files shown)


model/AssessmentResultsService.php
```diff
@@ -127,8 +145,7 @@ public function getDeliveryData(\taoDelivery_models_classes_execution_DeliveryEx
      */
     public function getPrintableRubric(\taoDelivery_models_classes_execution_DeliveryExecution $deliveryExecution)
     {
-        /** @var TestSessionService $deliveryService */
-        $testSessionService = $this->getServiceManager()->get(TestSessionService::SERVICE_ID);
+        $testSessionService = $this->getTestSessionService();
```
hutnikau: Why don't you get service using `TestSessionService::singleton()` here?

jsconan: Yes I could do that. I took the same behavior as [`DeliveryExecutionStateService`](https://github.com/oat-sa/extension-tao-proctoring/blob/fix/reporting/model/implementation/DeliveryExecutionStateService.php#L307) to ease the access to this service for potential reuse elsewhere in the class.

I can remove it and just address the singleton if you prefer this way.

hutnikau: Yes i think `getTestSessionService()` is redundant there.

jsconan:
Ok changes made.

PR #96, opened 2016-02-26
Tao 2000 use delivery log to save reports

hutnikau:


krampstudio:
@jbout could you please have a quick look, then drop the PR to someone else ?

PR #107, opened 2016-03-17
Tao 2002 auto termination after pause

Alroniks:
https://oat-sa.atlassian.net/browse/TAO-2002

Related to https://github.com/oat-sa/extension-tao-act/pull/186

### DIFF
--- a/model/implementation/DeliveryExecutionStateService.php
+++ b/model/implementation/DeliveryExecutionStateService.php
@@ -20,11 +20,15 @@
 
 namespace oat\taoProctoring\model\implementation;
 
+use DateInterval;
+use DateTime;
+use DateTimeImmutable;
 use oat\oatbox\service\ConfigurableService;
-use oat\taoQtiTest\models\TestSessionMetaData;
+use oat\taoAct\model\testSession\TestSessionProcessor;
 use oat\taoDelivery\models\classes\execution\DeliveryExecution;
 use oat\oatbox\event\EventManager;
 use oat\taoProctoring\model\event\DeliveryExecutionTerminated;
+use oat\taoQtiTest\models\TestSessionMetaData;
 
 /**
  * Class DeliveryExecutionStateService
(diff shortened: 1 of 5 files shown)


Alroniks:
@krampstudio Please, take a look also.

model/implementation/DeliveryExecutionStateService.php
```diff
@@ -88,6 +92,32 @@ public function waitExecution(DeliveryExecution $deliveryExecution)
...
+                if ($wasPausedAt && $this->hasOption('termination_delay_after_pause')) {
+                    $delay = $this->getOption('termination_delay_after_pause');
+                    if ($wasPausedAt->add(new DateInterval($delay)) < (new DateTime())) {
+                        $this->terminateExecution($deliveryExecution, null, ['test' => TestSessionProcessor::TEST_CODE_INCOMPLETE]);
+
+                        return false;
+                    }
+                }
+            }
+
```
hutnikau: Could you move all this stuff from `waitExecution()` function to separate protected method (something like `isExpired()`)?

krampstudio:
Could you just explain me your approach. The goal is terminate executions paused from some amout of time, so what does trigger the given event ?

model/implementation/DeliveryExecutionStateService.php
```diff
@@ -20,11 +20,15 @@
 
 namespace oat\taoProctoring\model\implementation;
 
+use DateInterval;
+use DateTime;
+use DateTimeImmutable;
 use oat\oatbox\service\ConfigurableService;
-use oat\taoQtiTest\models\TestSessionMetaData;
+use oat\taoAct\model\testSession\TestSessionProcessor;
```
boajer: Is it necessary to introduce new dependency from act extension?

krampstudio: Good catch!

hutnikau: My opinion is that it is not necessary to add new dependency and use `TestSessionProcessor` here. You should not pass exit codes to the `DeliveryExecutionTerminated` event. You can implement new custom event (something like `DeliveryExecutionExpired`) and add listener of this event to the `TestSessionProcessor`. The reason of creation `TestSessionProcessor` was to let him decide for themselves which exit code should be saved, and encapsulate all work with exit codes in it.

Alroniks:
Closed per favor https://github.com/oat-sa/extension-tao-proctoring/pull/108

PR #117, opened 2016-04-05
Upgrade to 1.9.1 to update workstation & testCenter column to nullable

siwane:


jsconan:
Works fine, but you have forgotten to update the installer too.

PR #121, opened 2016-04-18
Fix/tao 2545 timer end

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2545

Requires: https://github.com/oat-sa/extension-tao-testqti/pull/436

Set the user uri on the test session if needed, to allow correct timer close.


jsconan:
@antoinerobin please do the code review then assign to @krampstudio for tests

PR #125, opened 2016-04-26
Sorting by last name for test takers in assignments to a delivery 

zagovorichev:
(by Proctor)
https://oat-sa.atlassian.net/browse/TAO-2555


antoinerobin:
it sorts only already assigned test taker and not the ones available for assignment

Alroniks:
@antoinerobin Rebased and fixed sorting for available list of test tackers

PR #128, opened 2016-05-02
Added sorting for delivery list for proctor

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-2556


Alroniks:
Hi @antoinerobin, please check it or assign to somebody else.

PR #132, opened 2016-05-13
add offline/online status

hutnikau:
depends on https://github.com/oat-sa/extension-tao-act/pull/225


helpers/DeliveryHelper.php
```diff
@@ -33,6 +33,7 @@
 use oat\taoProctoring\model\monitorCache\implementation\DeliveryMonitoringService;
 use oat\taoQtiTest\models\event\QtiTestChangeEvent;
 use qtism\runtime\tests\AssessmentTestSessionState;
+use oat\taoAct\model\runner\HeartbeatService;
```
jsconan: Please do not directly use the heartbeat here, as it is a customer specific. You should rather use an abstraction to handle the last activity. Thus an implementation of this abstraction can be done to handle the heartbeat, and this implementation should be registered from a customer extension.

helpers/DeliveryHelper.php
```diff
@@ -455,68 +456,11 @@ private static function adjustDeliveryExecutions($deliveryExecutions, $options)
...
-                    )
-                ),
-            );
-            /********/
-
             $executions = array();
             
-            
             $deliveryMonitoringService = ServiceManager::getServiceManager()->get(DeliveryMonitoringService::CONFIG_ID);
+            $heartbeatService = ServiceManager::getServiceManager()->get(HeartbeatService::SERVICE_ID);
```
jsconan: You should use an abstraction, not the heartbeat itself.

views/js/controller/Delivery/monitoring.js
```diff
@@ -426,13 +426,20 @@ define([
                 id: 'status',
                 label: __('Status'),
                 transform: function(value, row) {
-                    if(row && row.state && row.state.status){
-                        var status = _status.getStatusByCode(row.state.status);
-                        if(status){
-                            return status.label;
+                    var result = '',
+                        status,
+                        online = row.online ? 'online' : 'offline';
```
jsconan: This label should be translated

hutnikau:
@jsconan the branch has been updated according to your comments. Please review it again.

hutnikau:
Please note that it depends on https://github.com/oat-sa/extension-tao-act/pull/227

manifest.php
```diff
@@ -24,7 +24,7 @@
 	'label' => 'Proctoring',
 	'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '1.12.2',
+    'version' => '1.12.3',
```
jsconan: Not sure it is a fix! It seems rather to be a feature...

hutnikau: Fixed!

PR #138, opened 2016-05-20
Add checking of delivery execution status

hutnikau:
To reduce amount of queries to the DB and improve performance I added checking of connectivity status only for active delivery executions because there is no need to check this status for finished, paused or terminated executions.


hutnikau:
Logic of getting this value will be changed by Micheal soon.

PR #152, opened 2016-06-09
add js check and remove modernizr

krampstudio:
Add the js and feature check on the proctoring template.
Also remove modernizr as well.

requires https://github.com/oat-sa/tao-core/pull/853


jsconan:
Works fine in admin, but breaks in delivery

PR #155, opened 2016-06-20
Only revoke proctor authorization from existing deliveries

jsconan:


krampstudio:
It seems to work on my env.

PR #157, opened 2016-06-22
Use the position in the section instead of the whole test

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2694

Build the assessment progress status from the current position inside the section rather than the position inside the whole test.

### DIFF
--- a/model/monitorCache/implementation/DeliveryMonitoringData.php
+++ b/model/monitorCache/implementation/DeliveryMonitoringData.php
@@ -202,38 +202,12 @@ class DeliveryMonitoringData implements DeliveryMonitoringDataInterface
         if ($session !== null) {
             if ($session->isRunning()) {
                 $route = $session->getRoute();
-                $routeItems = $route->getAllRouteItems();
-                $currentPosition = $route->getPosition();
                 $currentSection = $session->getCurrentAssessmentSection();
-                $currentSectionId = $currentSection->getIdentifier();
+                $sectionItems = $route->getRouteItemsByAssessmentSection($currentSection);
+                $currentItem = $route->current();
+                $positionInSection = array_search($currentItem, $sectionItems->getArrayCopy(true));
 
-                $sectionCount = 0;
-                $positionInSection = 0;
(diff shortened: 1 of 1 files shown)


model/monitorCache/implementation/DeliveryMonitoringData.php
```diff
@@ -200,12 +200,40 @@ private function getProgress()
...
+
+                $sectionCount = 0;
+                $positionInSection = 0;
+                $offset = 0;
+                $offsetSection = 0;
+                $lastSection = null;
+
+                foreach ($routeItems as $routeItem) {
+                    $sections = $routeItem->getAssessmentSections();
+                    $sectionId = key(current($sections));
```
boajer: Hi @jsconan,  
May be it will be safer to do something like `$sections->key()` here? 
As far as I now using `array` related functions leads to different behaviour, so you are getting first object property and passing that value to `key()` function. Hopefully here we have the only property as for now that contains collected values itself  so it works, but its never call `Iterators` methods behind. 
BTW: such construction works differently in PHP7

model/monitorCache/implementation/DeliveryMonitoringData.php
```diff
@@ -200,12 +200,40 @@ private function getProgress()
...
 
         if ($session !== null) {
-            $pos = $session->getRoute()->getPosition();
-            $count = $session->getRouteCount();
-
             if ($session->isRunning()) {
-                $section = $session->getCurrentAssessmentSection();
-                $result =  __('%1$s - item %2$s/%3$s', $section->getTitle(), $pos+1, $count);
+                $route = $session->getRoute();
+                $routeItems = $route->getAllRouteItems();
```
hutnikau: I think that instead of getting all items and loop through them is better to get item from current section using `Route::getRouteItemsByAssessmentSection()`:
https://github.com/oat-sa/qti-sdk/blob/legacy/qtism/runtime/tests/Route.php#L972

boajer:
Updated, please review it again.

PR #158, opened 2016-06-23
in bulkActionPopup was added option reasonRequired

zagovorichev:
all fields in the form are required

https://oat-sa.atlassian.net/browse/TAO-2696


zagovorichev:
depends on https://github.com/oat-sa/tao-core/pull/867

PR #160, opened 2016-07-01
change the authorization service

krampstudio:
Requires https://github.com/oat-sa/tao-core/pull/882 
Requires https://github.com/oat-sa/extension-tao-delivery/pull/180

- refactor the `eligibilityEditor`
- create the `eligibilitTable` component to support the _byPassProctor_ property
- update the controller to use the new components
- implement the delivery authorization service to switch between proctored and un-proctored 
- Change the proctoring API to use the new authorization provider
- There's still a bug : a non-proctored delivery needs to be launched twice because the `offline autopause` plugin add a flag on the delivery a paused. This flag is reset into the proctoring's awaiting page, and obviously using a non proctored delivery this page is not executed...


views/js/component/eligibilityEditor.js
```diff
@@ -17,202 +17,232 @@
...
+             * Add eligibilities
+             * @param {jQueryElement} $container - where to append the component
+             * @returns {eligibilityEditor} chains the component
+             * @fires eligibilityEditor#ok with the selected eligibities in parameter
+             * @fires eligibilityEditor#cancel
+             */
+            add : function add($container){
+                return this.on('render', function(){
+                    var self = this;
+                    debugger;
```
ssipasseuth: debugger

PR #162, opened 2016-07-04
hotfix in case properties is not defined

llecaque:
@krampstudio and @ssipasseuth Could you just confirmed that this hotix is not against the idea of the feature ?


krampstudio:
The fix is ok.

PR #164, opened 2016-07-08
Delivery can be checked against browser in order to run

boajer:


jbout:
moved to authorization service

PR #173, opened 2016-07-18
improve multiple values finding in monitoring cache

krampstudio:
_in progress_
Do not merge.

### DIFF
--- a/model/monitorCache/implementation/MonitorCacheService.php
+++ b/model/monitorCache/implementation/MonitorCacheService.php
@@ -90,10 +90,6 @@ class MonitorCacheService extends DeliveryMonitoringService
     {
         $whereClause = '';
 
-        if (is_array($condition) && count($condition) === 1 && is_array(current($condition))) {
-            $condition = current($condition);
-        }
-
         if (is_string($condition) && in_array(mb_strtoupper($condition), ['OR', 'AND'])) {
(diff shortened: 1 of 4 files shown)


model/monitorCache/implementation/MonitorCacheService.php
```diff
@@ -90,10 +90,6 @@ protected function prepareCondition($condition, &$parameters, &$selectClause)
     {
         $whereClause = '';
 
-        if (is_array($condition) && count($condition) === 1 && is_array(current($condition))) {
```
hutnikau: This condition is necessary there. In case if a criteria is something like that:
```php
$deliveryMonitoringService->find([
    ['status' => 'finished']
]);
```
Without this condition Unit tests are fails.

model/monitorCache/implementation/MonitorCacheService.php
```diff
@@ -114,6 +110,8 @@ protected function prepareCondition($condition, &$parameters, &$selectClause)
 
             if ($value === null) {
                 $op = 'IS NULL';
+            } else if(is_array($value)){
+                $op = 'IN (' . join(',', array_map(function(){ return '?'; }, $value)) . ')';
```
hutnikau: Please add unit test for this case.

PR #183, opened 2016-07-27
Feature/tao 2805 history service

antoinerobin:
use : https://github.com/oat-sa/extension-tao-act/pull/315 to test it

### DIFF
--- a/helpers/TestCenterHelper.php
+++ b/helpers/TestCenterHelper.php
@@ -28,6 +28,9 @@ use oat\taoProctoring\model\mock\WebServiceMock;
 use oat\taoProctoring\model\TestCenterService;
 use core_kernel_classes_Resource;
 use DateTime;
+use qtism\data\AssessmentTest;
+use qtism\data\storage\xml\XmlDocument;
+use qtism\runtime\tests\AssessmentTestSession;
 use tao_helpers_Date as DateHelper;
 use oat\tao\helpers\UserHelper;
 use oat\taoProctoring\model\implementation\DeliveryService;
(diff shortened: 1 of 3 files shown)


helpers/TestCenterHelper.php
```diff
@@ -239,153 +238,80 @@ public static function removeDiagnostic(core_kernel_classes_Resource $testCenter
...
+    public static function getSessionHistory($testCenter, $sessions, $logHistory = false, $options = array())
     {
         if (!is_array($sessions)) {
             $sessions = $sessions ? [$sessions] : [];
         }
+        
+        $deliveryLog = ServiceManager::getServiceManager()->get(DeliveryLog::SERVICE_ID);
+
+        $periodStart = null;
+        $periodEnd = null;
```
jsconan: These two variables are never used

helpers/TestCenterHelper.php
```diff
@@ -239,153 +238,80 @@ public static function removeDiagnostic(core_kernel_classes_Resource $testCenter
...
-                'context' => 'Section 1 - item 3/3'
-            ];
+            $valid = false;
+            $deliveryExecution = \taoDelivery_models_classes_execution_ServiceProxy::singleton()->getDeliveryExecution($sessionUri);
+            $delivery = $deliveryExecution->getDelivery();
+            $executions = $deliveryService->getCurrentDeliveryExecutions($delivery, $testCenter, $options);
+            foreach($executions as $execution){
+                if($execution['delivery_execution_id'] === $sessionUri && $execution['authorized_by'] === $proctor->getUri()){
+                    $valid = true;
+                    continue;
```
jsconan: Maybe a break instruction could be better here

helpers/TestCenterHelper.php
```diff
@@ -239,153 +238,80 @@ public static function removeDiagnostic(core_kernel_classes_Resource $testCenter
...
+            if($logHistory){
+                $deliveryLog->log($deliveryExecution->getIdentifier(), 'HISTORY', array());
+            }
+
+            $logs = $deliveryLog->get($deliveryExecution->getIdentifier());
+            $exportable = array();
+            foreach($logs as $data){
+                if($data['event_id'] !== 'HEARTBEAT' && strpos($data['event_id'], 'latency') === FALSE){
+                    $author = new \core_kernel_classes_Resource($data['created_by']);
+                    $proctorRole = new \core_kernel_classes_Resource('http://www.tao.lu/Ontologies/TAOProctor.rdf#ProctorRole');
```
jsconan: I don't known what is the performances impact of creating an instance of Resource, but maybe it could be better to factorise this variable and move this line outside of the loops...

helpers/TestCenterHelper.php
```diff
@@ -239,153 +238,80 @@ public static function removeDiagnostic(core_kernel_classes_Resource $testCenter
...
+                    if(is_string($data['data'])){
+                        $data['data'] = json_decode($data['data'], true);
+                    }
+                    if(isset($data['data']['type'])) {
+                        $event_id = $data['data']['type'];
+
+                        $context = (isset($data['data']['context']['readable']))?$data['data']['context']['readable'] : '';
+                        $details = (isset($data['data']['context']['shortcut']))?$data['data']['context']['shortcut']: '';
+                    } else {
+                        $details = (isset($data['data']['reason']['reasons']) && !is_null($data['data']['reason']['reasons']))?array_merge(array_values($data['data']['reason']['reasons']), array($data['data']['reason']['comment'])) : '';
```
jsconan: Is the 'reason' index always exist if the type is not provided ? This could lead to issue if the data is extended later. I would rather place a ` elseif (isset($data['data']['reason']))` here. Or whatever could ensure the needed format.

PR #184, opened 2016-07-27
Refactor/recurrent terminate paused assessment

hutnikau:


jbout:
This is not a suitable solution for recurring tasks, as it would cause havoc on all non cron-job implemented queues.

PR #191, opened 2016-08-02
Check heartbeats

hutnikau:


model/implementation/TestSessionService.php
```diff
@@ -129,7 +129,7 @@ public function isExpired(DeliveryExecution $deliveryExecution)
 
             if (
                 DeliveryExecution::STATE_PAUSED !== $executionState
-                || !$lastPauseEvent = current(array_reverse($deliveryLogService->get($deliveryExecution->getIdentifier())))
+                || !$lastPauseEvent = current(array_reverse($deliveryLogService->get($deliveryExecution->getIdentifier(), 'HEARTBEAT')))
```
krampstudio: as far as I know, the heartbeat is a feature from taoAct, not taoProctoring ?

hutnikau:
@boajer I decided find events by test-taker's identifier. I guess this is more reliable solution than rely on event ID. Please review it again.

PR #192, opened 2016-08-04
Fix/tao 2820 pause on disconnect

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2820

Requires: https://github.com/oat-sa/extension-tao-testqti/pull/539

Remove the persistent state cleaner (related to pause handling)


views/js/controller/DeliveryServer/awaiting.js
```diff
@@ -121,9 +114,13 @@ define([
...
              */
             function authorized(){
+                var box = list.getElement().find('.list');
                 loadingBar.stop();
-                //@todo it would be nice to smoothen the transition
-                $container.removeClass('authorization-in-progress');
+
+                box.fadeOut(250, function() {
+                    $container.removeClass('authorization-in-progress');
+                    box.fadeIn(250);
```
krampstudio: Maybe this is bikeshedding, but animations are in the CSS land, especially since you've a dedicated class. You should add a transition on the class `authorization-in-progress`.

PR #195, opened 2016-08-05
check delivery state

hutnikau:
Steps to reproduce:
1. create un-proctor delivery
2. run delivery as test taker
3. close browser tab
4. log in as test taker again. See this delivery in _In progress_ section. Don't close this tab
5. log in as proctor in another browser
6. terminate this delivery. See that status of this delivery is _terminated_
7. return back to the test-taker's window and press _Resume_ on  this delivery (but it should desappear from this list)
8. page refreshes but this delivery still in _In progress_ section
9. return to proctor's window and refresh page. See that this session has _in progress_ status (cannot be in _in progress_ state as it was terminated)
10. Try to terminate it. See error message.


krampstudio:
principle seems correct. @jsconan could you please take care of the merge ?

PR #203, opened 2016-08-23
use select request instead of fetching schema

hutnikau:
I found that getting schema takes very long time. After this change I got significant accelerations for such requests as heartbeat, move (next, prev, skip), submitting item data and other requests where delivery monitoring storage is used. 
For example I got 50% acceleration for heartbeat call (189ms vs 361ms).


hutnikau:
The branch has been updated. Please review it again.

PR #211, opened 2016-09-13
All executions retrieved for testtaker from KV storage

boajer:
depends on https://github.com/oat-sa/extension-tao-delivery/pull/195


jbout:
https://github.com/oat-sa/extension-tao-delivery/pull/195 has been replaced by https://github.com/oat-sa/extension-tao-delivery/pull/196

PR #220, opened 2016-09-27
add columns delivery id and test center id to monitoring table and us…

antoinerobin:
…e it to avoid useless and time consuming left join

### DIFF
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -507,6 +507,71 @@ class Updater extends common_ext_ExtensionUpdater {
         }
         $this->skip('3.6.6', '3.6.13');
 
+        if ($this->isVersion('3.6.13')) {
+
+            // create new column for test center and delivery
+            $persistenceId = $this->getServiceManager()->get(DeliveryMonitoringService::CONFIG_ID)->getOption(DeliveryMonitoringService::OPTION_PERSISTENCE);
+            $persistence = \common_persistence_Manager::getPersistence($persistenceId);
+            $schemaManager = $persistence->getDriver()->getSchemaManager();
+            $schema = $schemaManager->createSchema();
+            $fromSchema = clone $schema;
+            try {
+                $tableLog = $schema->getTable(DeliveryMonitoringService::TABLE_NAME);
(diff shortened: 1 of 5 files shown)


scripts/update/Updater.php
```diff
@@ -507,6 +507,71 @@ public function update($initialVersion) {
         }
         $this->skip('3.6.6', '3.6.13');
 
+        if ($this->isVersion('3.6.13')) {
+
+            // create new column for test center and delivery
+            $persistenceId = $this->getServiceManager()->get(DeliveryMonitoringService::CONFIG_ID)->getOption(DeliveryMonitoringService::OPTION_PERSISTENCE);
```
jbout: Update script can be costly, please provide a seperate script to migrate from one configuration to the other

antoinerobin: done, I just forgot to push it

scripts/update/Updater.php
```diff
@@ -507,6 +507,30 @@ public function update($initialVersion) {
...
+            $persistence = \common_persistence_Manager::getPersistence($persistenceId);
+            $schemaManager = $persistence->getDriver()->getSchemaManager();
+            $schema = $schemaManager->createSchema();
+            $fromSchema = clone $schema;
+            try {
+                $tableLog = $schema->getTable(DeliveryMonitoringService::TABLE_NAME);
+                $tableLog->addColumn(DeliveryMonitoringService::COLUMN_DELIVERY_ID, "string", array("notnull" => true, "length" => 255));
```
hutnikau: Column value cannot be null and this column will not be added to the table with existing data. Update script will fail with `column "delivery_id" contains null values` message.
Add default value to this column and then drop it:
```
$tableLog->addColumn(DeliveryMonitoringService::COLUMN_DELIVERY_ID, "string", array("notnull" => true, "length" => 255, 'default' => ''));
```
After column will be created call
```
ALTER TABLE delivery_monitoring ALTER COLUMN delivery_id DROP DEFAULT;
```
The same thing should be done for test center column

jbout:
test centers are in the process of being moved out to https://github.com/oat-sa/extension-tao-testcenter please migrate the improvement to this extension, from the branch refactor/split-testcenter-proctoring

PR #222, opened 2016-10-10
fix unit test

hutnikau:


antoinerobin:
this test works for me, ```$authorization``` has 4 entries

hutnikau:
How there can be 4 entries if there are only 3 proctors? Probably you have old generis version. `searchInstances()` method was broken there.

PR #229, opened 2016-11-08
Authorize and Pause buttons moved to the separate columns

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-3308
https://oat-sa.atlassian.net/browse/TAO-3309

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -25,7 +25,7 @@ return array(
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '3.6.16',
+    'version' => '3.6.17',
     'author' => 'Open Assessment Technologies SA',
     'requires' => array(
         'tao' => '>=4.5.0',
--- a/views/js/controller/Delivery/monitoring.js
+++ b/views/js/controller/Delivery/monitoring.js
@@ -33,6 +33,7 @@ define([
     'taoProctoring/helper/status',
     'tpl!taoProctoring/templates/delivery/deliveryLink',
     'tpl!taoProctoring/templates/delivery/statusFilter',
+    'tpl!taoProctoring/templates/delivery/actionBtn',
     'ui/datatable',
     'select2'
 ], function (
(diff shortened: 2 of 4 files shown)


views/js/controller/Delivery/monitoring.js
```diff
@@ -560,6 +528,50 @@ define([
...
+                transform: function(value, row) {
+                    var status;
+                    var inactive = true;
+
+                    if (row.state && row.state.status){
+                        status = _status.getStatusByCode(row.state.status);
+                        inactive = !status || status.can.authorize !== true;
+                    }
+
+                    return actionBtnTpl({
```
ssipasseuth: I was actually expected to have this action button as part of a configurable option for the datatable like the last column action... so modifying the datatable component to allow inserting action not only in the end but also the location we want. This way we don't have to redefine views/js/templates/delivery/actionBtn.tpl each time and add it from the  transform() callback.
I am not sure it is easily feasible but since you are ahead of schedule, what about exploring this option ?

manifest.php
```diff
@@ -25,7 +25,7 @@
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '3.6.16',
+    'version' => '3.6.17',
```
ssipasseuth: I believe this is more an improvement so an intermediate version increase to 3.7 makes more sense, no ?

views/js/controller/Delivery/monitoring.js
```diff
@@ -560,6 +528,50 @@ define([
...
+                        icon: 'pause',
+                        type: 'pause',
+                        title: __('Pause session'),
+                        disabled: inactive ? 'disabled' : ''
+                    });
+                }
+            });
```
ssipasseuth: we add a type to each column, if it is of an action type, it support an action callback or an array of callback. this way we generalize the action column concept so it can be anywhere instead of having to be always at the end.
 (The current option action should be preserved however for retrocompability.)

PR #230, opened 2016-11-08
As a proctor, I want the Actions column to be reworded as Administrat…

zagovorichev:
…ion in the proctor all sessions screen.
https://oat-sa.atlassian.net/browse/TAO-3305

### DIFF
--- a/views/js/controller/Delivery/monitoring.js
+++ b/views/js/controller/Delivery/monitoring.js
@@ -599,6 +599,9 @@ define([
                         report : $list.find('.action-bar').children('.tool-irregularity')
                     });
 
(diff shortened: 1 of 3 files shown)


views/js/controller/Delivery/monitoring.js
```diff
@@ -599,6 +599,9 @@ define([
                         report : $list.find('.action-bar').children('.tool-irregularity')
                     });
 
+                    // rewrite column name
+                    $('th.actions', $list).text(__('Administration'));
```
ssipasseuth: can you make it a configurable option for the core datatable ?

PR #231, opened 2016-11-08
As a proctor, I want test takers waiting to be authorized to be highl…

zagovorichev:
…ighted when monitoring a delivery for better visibility.

https://oat-sa.atlassian.net/browse/TAO-3306


views/js/controller/Delivery/monitoring.js
```diff
@@ -554,6 +556,10 @@ define([
                             if (row.state.status === 'INPROGRESS') {
                                 result = status.label;
                             }
+                            if (result === 'Awaiting') {
+                                // highlight row
+                                result = highlightedStatusTpl({status: result});
+                            }
```
ssipasseuth: I have the same comments as for https://github.com/oat-sa/extension-tao-proctoring/pull/229: although this solution may works, what about exploring the possibility to extend the core datatable component to allow adding a class to a row under specific case like in this example. It will indeed be better if the highlight is not dependent from the need to manually tweak the cell with views/js/templates/delivery/highlightedStatus.tpl then search and set the line class in l.608.

manifest.php
```diff
@@ -25,7 +25,7 @@
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '3.6.17',
+    'version' => '3.6.18',
```
ssipasseuth: a feature or a bug?

zagovorichev:
depends on https://github.com/oat-sa/tao-core/pull/1063

@ssipasseuth Check it, please

views/js/controller/Delivery/monitoring.js
```diff
@@ -554,6 +555,9 @@ define([
                             if (row.state.status === 'INPROGRESS') {
                                 result = status.label;
                             }
+                            if (result === 'Awaiting') {
+                                highlightRows.push(row.id);
```
ssipasseuth: you cannot highlight the row directly here ?

ssipasseuth:
(I merged develop into this branch already)

views/js/controller/Delivery/monitoring.js
zagovorichev: no, because transform worked before rendering, so row is not exists

ssipasseuth: ok fair enough

PR #232, opened 2016-11-09
As a proctor, I want the Print button to respond with a more detailed…

zagovorichev:
… message when there are no reports to print.

https://oat-sa.atlassian.net/browse/TAO-3299


views/js/templates/delivery/reportExceptedDeliveries.tpl
```diff
@@ -0,0 +1,7 @@
+<p>{{__ "If you continue, these deliveries will be missing from the report."}}</p>
```
ssipasseuth: Would be useful to state the reason why to the user

controller/Reporting.php
```diff
@@ -208,4 +217,47 @@ public function reports(){
...
+
+        $idList = $id;
+        if (!is_array($idList)) {
+            $idList = [$idList];
+        }
+
+        /** @var $assessmentResultsService \oat\taoProctoring\model\AssessmentResultsService */
+        $assessmentResultsService = $this->getServiceManager()->get('taoProctoring/AssessmentResults');
+
+        $excepted = [];
```
ssipasseuth: please use a more meaningful variable names so the next developer understand the meaning of your code (this comment is also applicable to other similar variables and file names: excepted, allExcepted, reportExceptedDeliveries, reportExceptedDeliveries.tpl, excepted etc.)

zagovorichev:
depends on https://github.com/oat-sa/tao-core/pull/1062

@ssipasseuth Check it, please

PR #234, opened 2016-11-11
Tao 3300 detailed session history button

hutnikau:


config/default/TestSessionServiceHistory.conf.php
```diff
@@ -0,0 +1,22 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2016 (original work) Open Assessment Technologies SA;
+ *
+ *
+ */
+
+return new oat\taoProctoring\model\implementation\TestSessionHistoryService();
```
antoinerobin: you file should be named ```TestSessionHistoryService.conf.php```

PR #237, opened 2016-11-15
Feature/text converter

siwane:
Require https://github.com/oat-sa/tao-core/pull/1064

Use service to convert text contents. A nullTextConverter service is used because to keep value.


controller/TestCenter.php
```diff
@@ -67,7 +70,7 @@ public function testCenter()
         $testCenter  = $this->getCurrentTestCenter();
         $data = array(
             'testCenter' => $testCenter->getUri(),
-            'title' => __('Test site %s', $testCenter->getLabel()),
+            'title' => sprintf($this->convert('Test site %s'), $testCenter->getLabel()),
```
antoinerobin: how will it be translatable ?

controller/TestCenterManager.php
```diff
@@ -69,28 +74,28 @@ public function editCenter()
                 $testCenter = $binder->bind($myForm->getValues());
 
                 $this->setData("selectNode", \tao_helpers_Uri::encode($testCenter->getUri()));
-                $this->setData('message', __('Test center saved'));
+                $this->setData('message', $this->convert('Test center saved'));
```
antoinerobin: no entry in text registry

controller/TestCenterManager.php
```diff
@@ -69,28 +74,28 @@ public function editCenter()
...
         $this->setData('administratorForm', $administratorForm->render());
 
         $proctorProperty = new \core_kernel_classes_Property(ProctorManagementService::PROPERTY_ASSIGNED_PROCTOR_URI);
         $proctorForm = \tao_helpers_form_GenerisTreeForm::buildReverseTree($testCenter, $proctorProperty);
-        $proctorForm->setData('title', __('Assign proctors'));
+        $proctorForm->setData('title', $this->convert('Assign proctors'));
         $this->setData('proctorForm', $proctorForm->render());
 
-        $this->setData('formTitle', __('Edit test center'));
+        $this->setData('formTitle', $this->convert('Edit test center'));
```
antoinerobin: no entry in text registry

PR #247, opened 2016-11-21
Authorize and Pause columns ordering were fixed

zagovorichev:
TAO-3309


zagovorichev:
@ssipasseuth Wait with a review, please, I need to implement this: If possible the button should be center justified within the column.

PR #255, opened 2016-11-30
Added message for the singular selection

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-3299


ssipasseuth:
Hi, just tested it and it works when selecting only one row then clicking on print score on the top toolbar.
However I still have the plural version when I directly click on the print score button from the administration column.

zagovorichev:
@ssipasseuth Fixed

ssipasseuth:
ok, it works now!

PR #270, opened 2017-01-30
Tao 3735 pause active test sesions

hutnikau:


views/js/controller/Tools/pauseActiveExecutions.js
```diff
@@ -0,0 +1,22 @@
+define(['jquery', 'helpers', 'util/url', 'ui/feedback'], function($, helpers, url, feedback){
```
jsconan: You should add the license header.
Also the `'use strict';` instruction is missing.

hutnikau:
@jsconan i have pushed commit with changed you have requested. Please review it again.

PR #277, opened 2017-02-10
Tao 3786 start time filter

hutnikau:


scripts/update/Updater.php
```diff
@@ -165,5 +165,32 @@ public function update($initialVersion)
         }
 
         $this->skip('4.2.0', '4.3.1');
+
+        if ($this->isVersion('4.3.1')) {
+            /** @var DeliveryMonitoringService $monitoring */
```
antoinerobin: It can take some times so it should be an external script

hutnikau:
@antoinerobin I've made necessary changes. Please review it again

PR #288, opened 2017-02-22
Fix/tao 3788 test taker time

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3788

Use the right time target when updating the cache of the delivery execution with the remaining time.


jsconan:
Code will be moved in taoQtiTest\TestSession for more consistency

PR #291, opened 2017-02-24
Tao 3862 test taker cancel authorization request

hutnikau:


model/implementation/DeliveryExecutionStateService.php
```diff
@@ -50,6 +49,16 @@ class DeliveryExecutionStateService extends ConfigurableService implements \oat\
     private $testSessionService;
 
     /**
+     * @inheritdoc
+     * @param DeliveryExecutionInterface $deliveryExecution
+     * @param $state
+     */
+    public function setState(DeliveryExecutionInterface $deliveryExecution, $state)
+    {
+        $deliveryExecution->setState($state);
```
hutnikau: This is temporary default implementation.
Logic of changing state i will implement in further PRs.

jbout:
Looks good, but not yet tested.

PR #302, opened 2017-03-13
Fix/tao 3944 back button

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3944

On the sessions history page, add a button to return to the list of delivery executions.
This button button should also work in the context of LTI.
For this reason the patch is a little bit more complex than expected as we need to take care of the actual route to reach, that may not be in the same extension.

The service `TestSessionHistoryService` has been updated to provide the right URL for the history page and the back URL.


jsconan:
@antoinerobin please validate the PHP code
@krampstudio please review the frontend side

krampstudio:
Waiting @antoinerobin approval to merge

PR #303, opened 2017-03-13
Feature/tao 3940 display delivery

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3940

Always displays the delivery name in the list of delivery executions.
Make this column sortable.


jsconan:
Closed to fix conflicts in a new PR

PR #332, opened 2017-04-05
Tao 4136 log delivery execution pause

hutnikau:
depends on:
https://github.com/oat-sa/extension-tao-testqti/pull/796/files
https://github.com/oat-sa/extension-tao-test-runner-plugins/pull/4/files


hutnikau:
@antoinerobin i have fixed double log. Please test it again.

PR #333, opened 2017-04-07
Cretae delivery activity datatable

hutnikau:


scripts/update/Updater.php
```diff
@@ -273,7 +273,7 @@ public function update($initialVersion)
                 [DeliveryExecutionStateService::SERVICE_ID, 'catchSessionPause']
             );
             $this->getServiceManager()->register(EventManager::SERVICE_ID, $eventManager);
-            $this->setVersion('4.15.0');
+            $this->setVersion('4.16.0');
```
boajer: Are you sure ?

PR #336, opened 2017-04-10
Fix/tao 4128 dont forward authorized deliveries to execution

eacraft13:
Notes:

- The short polling frequency was reduced to 1 second. This prevents test takers from waiting 10 seconds before proceeding to test after a browser refresh on an authorized test. Please let me know any opinions on this. I understand we need to account for limited connectivity, but I don't believe a ~400B response will be impactful, even at 1 second intervals. Another option would be to keep a 10 second poll timeout, but update `core/polling` to immediately call the passed action instead of waiting for the timeout. Perhaps @ssipasseuth can comment on the best path for this.

- There is still a path to directly launching an authorized delivery. If a delivery is authorized and the test taker launches test from lti (as opposed to refreshing the screen), they will go directly to the delivery (instead of the 'awaitingAuthorization' page). This is due to the nature of the [taoDelivery extension](https://github.com/oat-sa/extension-tao-delivery/blob/master/controller/DeliveryServer.php#L187). The try/catch block will forward non-authorized deliveries to the 'awaitingAuthorization' page. I think updating the flow of taoDelivery is outside the scope of this ticket and thus will leave this shortcoming intact.

### DIFF
--- a/controller/DeliveryServer.php
+++ b/controller/DeliveryServer.php
@@ -102,11 +102,6 @@ class DeliveryServer extends DefaultDeliveryServer
 
         $runDeliveryUrl = _url('runDeliveryExecution', null, null, array('deliveryExecution' => $deliveryExecution->getIdentifier()));
 
-        // if the test taker is already authorized, straight forward to the execution
-        if (DeliveryExecutionState::STATE_AUTHORIZED === $executionState) {
-            return $this->redirect($runDeliveryUrl);
-        }
-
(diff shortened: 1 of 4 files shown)


eacraft13:
Great. `polling.next()` is perfect. I reverted back to 10s polling as well.

jsconan:
Tom is ok to address the remaining issue you discovered in another ticket. So I am going to merge the PR right now.

PR #343, opened 2017-04-21
Tao 4233 log finishing of delivery execution

hutnikau:


model/implementation/DeliveryExecutionStateService.php
```diff
@@ -266,6 +267,8 @@ public function finishExecution(DeliveryExecution $deliveryExecution, $reason =
      */
     public function finish(DeliveryExecution $deliveryExecution, $reason = null)
     {
+        $eventManager = $this->getServiceManager()->get(EventManager::SERVICE_ID);
+        $eventManager->trigger(new DeliveryExecutionFinished($deliveryExecution));
```
boajer: Maybe you can consider event triggering after state has been actually successfully changed?

PR #347, opened 2017-04-24
Feature/tao 4135 configurable pause visibility

antoinerobin:


antoinerobin:
Yes, the terminate should no be hidden.
I modified it

PR #352, opened 2017-04-25
Feature/tao 4232 exit button

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-4232

Add a button to exit the proctor monitoring page when the `exitUrl` config is set.


jsconan:
I added the feature the same way the "Back to session button" was made for the reporting screen. As I was wondering if the button was only related to LTI, I first added it from the proctoring screen.

However as it seems there is no more need from this button outside of the LTI mode, I will close this PR and make the feature as you recommend it.

jsconan:
I reopen the PR because the acceptance criteria need a confirm message, that can only be achieved from the controller of taoProctoring. Also if the branding is not enabled, the exit button won't be displayed.

hutnikau:
I just tried to disable branding and exit button has been successfully rendered.
You can extend taoProctoring js controller for example in this way:
https://github.com/oat-sa/extension-lti-proctoring/compare/develop...experimental/lti_specific_controller?expand=1#diff-d4f049018ec1c3723f17bfa3c047745aR29

This is LTI specific feature and we don't have to change taoProctoring for just to render this button.

jsconan:
Ok, I found another solution. Not by overriding the controller, but by adding an extra controller on LTI.

PR #356, opened 2017-04-28
TAO-4236 Assessment Activity Auto Refresh

eacraft13:
[TAO-4236](https://oat-sa.atlassian.net/browse/TAO-4236)

### DIFF
--- a/config/default/ActivityMonitoringService.conf.php
+++ b/config/default/ActivityMonitoringService.conf.php
@@ -9,5 +9,6 @@ use oat\taoProctoring\model\ActivityMonitoringService;
 
 return new ActivityMonitoringService([
     ActivityMonitoringService::OPTION_ACTIVE_USER_THRESHOLD => 300,
-    ActivityMonitoringService::OPTION_AUTO_REFRESH_INTERVAL => 60,
+    ActivityMonitoringService::OPTION_COMPLETED_ASSESSMENTS_AUTO_REFRESH => 30,
+    ActivityMonitoringService::OPTION_ASSESSMENT_ACTIVITY_AUTO_REFRESH => 60,
 ]);
--- a/config/default/ActivityMonitoringService.conf.php
+++ b/config/default/ActivityMonitoringService.conf.php
@@ -9,6 +9,5 @@ use oat\taoProctoring\model\ActivityMonitoringService;
 
 return new ActivityMonitoringService([
     ActivityMonitoringService::OPTION_ACTIVE_USER_THRESHOLD => 300,
-    ActivityMonitoringService::OPTION_COMPLETED_ASSESSMENTS_AUTO_REFRESH => 30,
-    ActivityMonitoringService::AUTO_REFRESH_INTERVAL => 60,
+    ActivityMonitoringService::OPTION_AUTO_REFRESH_INTERVAL => 60,
 ]);
--- a/config/default/ActivityMonitoringService.conf.php
+++ b/config/default/ActivityMonitoringService.conf.php
@@ -10,4 +10,5 @@ use oat\taoProctoring\model\ActivityMonitoringService;
 return new ActivityMonitoringService([
     ActivityMonitoringService::OPTION_ACTIVE_USER_THRESHOLD => 300,
     ActivityMonitoringService::OPTION_COMPLETED_ASSESSMENTS_AUTO_REFRESH => 30,
+    ActivityMonitoringService::AUTO_REFRESH_INTERVAL => 60,
 ]);
--- a/controller/Tools.php
+++ b/controller/Tools.php
@@ -46,14 +46,15 @@ class Tools extends SimplePageModule
     {
         $service = $this->getServiceManager()->get(ActivityMonitoringService::SERVICE_ID);
 
-        // Data
-        $this->setData('activity_data', $service->getData());
+        // Reason categories
         $this->setData('reason_categories', DeliveryHelper::getAllReasonsCategories());
 
         // Config
         $this->setData('config', [
-            ActivityMonitoringService::OPTION_AUTO_REFRESH_INTERVAL =>
-                $service->getOption(ActivityMonitoringService::OPTION_AUTO_REFRESH_INTERVAL),
+            ActivityMonitoringService::OPTION_ASSESSMENT_ACTIVITY_AUTO_REFRESH =>
+                $service->getOption(ActivityMonitoringService::OPTION_ASSESSMENT_ACTIVITY_AUTO_REFRESH),
+            ActivityMonitoringService::OPTION_COMPLETED_ASSESSMENTS_AUTO_REFRESH =>
+                $service->getOption(ActivityMonitoringService::OPTION_COMPLETED_ASSESSMENTS_AUTO_REFRESH),
         ]);
 
         $this->setView('Tools/assessment_activity.tpl');
--- a/controller/Tools.php
+++ b/controller/Tools.php
@@ -87,13 +87,13 @@ class Tools extends SimplePageModule
      */
     public function completedAssessmentsData()
     {
-        $param = $this->getRequestParameter('interval');
+        $timePeriod = $this->getRequestParameter('interval');
 
         $eventLog = $this->getServiceManager()->get(\oat\taoEventLog\model\LoggerService::SERVICE_ID);
 
         $tz = new \DateTimeZone(\common_session_SessionManager::getSession()->getTimeZone());
-        $timeKeys = $this->getTimeKeys($param);
-        $interval = $this->getInterval($param);
+        $timeKeys = $this->getTimeKeys($timePeriod);
+        $interval = $this->getInterval($timePeriod);
 
         foreach ($timeKeys as $timeKey) {
             $to = clone($timeKey);
@@ -147,12 +147,12 @@ class Tools extends SimplePageModule
     /**
      * @return \DateInterval
      */
-    private function getInterval($param)
+    private function getInterval($timePeriod)
     {
         $interval = new \DateInterval('PT1H');
 
-        if ($param) {
-            switch ($param) {
+        if ($timePeriod) {
+            switch ($timePeriod) {
                 case 'day':
                     $interval = new \DateInterval('PT1H');
                     break;
@@ -177,7 +177,7 @@ class Tools extends SimplePageModule
     /**
      * @return \DateTime[]
      */
-    private function getTimeKeys($param)
+    private function getTimeKeys($timePeriod)
     {
         /** @var ActivityMonitoringService $service */
         $service = $this->getServiceManager()->get(ActivityMonitoringService::SERVICE_ID);
@@ -185,8 +185,8 @@ class Tools extends SimplePageModule
         $amount = null;
         $startDate = null;
 
-        if ($param) {
-            switch ($param) {
+        if ($timePeriod) {
+            switch ($timePeriod) {
                 case 'day':
                     $startDate = new \DateTime('now', new \DateTimeZone('UTC'));
                     break;
@@ -209,6 +209,6 @@ class Tools extends SimplePageModule
             }
         }
 
-        return $service->getTimeKeys($this->getInterval($param), $startDate, $amount);
+        return $service->getTimeKeys($this->getInterval($timePeriod), $startDate, $amount);
     }
 }
--- a/controller/Tools.php
+++ b/controller/Tools.php
@@ -67,8 +67,6 @@ class Tools extends SimplePageModule
         $service = $this->getServiceManager()->get(ActivityMonitoringService::SERVICE_ID);
         $data = $service->getData();
 
-        // todo - optimize these calls (only need one call to getData())
-
         // Assessment Activity Data
         $assessmentActivityData = $data;
 
--- a/controller/Tools.php
+++ b/controller/Tools.php
@@ -45,13 +45,34 @@ class Tools extends SimplePageModule
     public function assessmentActivity()
     {
         $service = $this->getServiceManager()->get(ActivityMonitoringService::SERVICE_ID);
+
+        // Data
         $this->setData('activity_data', $service->getData());
         $this->setData('reasonCategories', DeliveryHelper::getAllReasonsCategories());
+
+        // Config
         $this->setData('completed_assessments_config', [
             ActivityMonitoringService::OPTION_COMPLETED_ASSESSMENTS_AUTO_REFRESH =>
                 $service->getOption(ActivityMonitoringService::OPTION_COMPLETED_ASSESSMENTS_AUTO_REFRESH),
         ]);
+        $this->setData('assessment_activity_config', [
+            ActivityMonitoringService::AUTO_REFRESH_INTERVAL =>
+                $service->getOption(ActivityMonitoringService::AUTO_REFRESH_INTERVAL),
+        ]);
+
         $this->setView('Tools/assessment_activity.tpl');
+    }
+
+    /**
+     * Show assessment activity data as json
+     */
+    public function assessmentActivityData()
+    {
+        $service = $this->getServiceManager()->get(ActivityMonitoringService::SERVICE_ID);
+        $this->returnJson([
+            'success' => true,
+            'data' => $service->getData()
+        ]);
     }
 
     /**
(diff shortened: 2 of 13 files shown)


views/js/controller/Tools/assessmentActivity.js
```diff
@@ -16,187 +16,115 @@
...
+    var interval;
+
+    function handleError(jqXHR, textStatus, errorThrown) {
+        var res = JSON.parse(jqXHR.responseText);
+        feedback().error(res.message || errorThrown || 'Unknown error occurred');
+    }
 
-    function doPause(reason) {
+    function getAssessmentActivity(callback) {
         $.ajax({
```
ssipasseuth: what about using core/dataProvider/request instead ?

views/js/controller/Tools/assessmentActivity/index.css
```diff
@@ -0,0 +1,7 @@
+.container.assessment-activity {
+    padding-left: 20px;
+}
+
+.container.assessment-activity button.pause-active-test-sessions {
+    background-color: red;
```
ssipasseuth: don't we have a color variable you may reuse here ?

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -16,187 +16,115 @@
...
-                                    text:  interval === 'day' ? __('Hours') : __('Days')
+                getCompletedAssessmentsConfig(function (completedAssessmentConfigRes) {
+
+                    if (!completedAssessmentConfigRes.success) {
+                        return handleError(completedAssessmentConfigRes.error);
+                    }
+
+                    // if (completedAssessmentConfigRes.data.completed_assessments_auto_refresh) {
+                    if (true) {
+                        interval = setInterval(function () {
```
ssipasseuth: please use the core/polling module for this.
It allow better manage the life cycle of the polling.

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -16,187 +16,115 @@
...
                                 }
-                            }
-                        }
+
+                                _.each(assessmentActivityRes.data, function (datum, key) {
+                                    $('.' + key.replace('_', '-'), $container).text(datum);
+                                });
+                            });
+                        // }, completedAssessmentConfigRes.data.completed_assessments_auto_refresh * 1000);
+                        }, 1000 * 2);
```
ssipasseuth: this need to be set in a server side configuration

eacraft13:
@hutnikau [commit c84682b3...](https://github.com/oat-sa/extension-tao-proctoring/pull/356/commits/c84682b3e723ca3e4d401d5c15689324e9677ed8) completely reorganizes the entire controller. This commit can be reverted if you don't approve. Or let me know if you have suggestions.

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -100,88 +158,89 @@ define([
...
+                        show: false
+                    }
+                }
+            })
+            .render();
+        } else {
+            activityGraph.refresh({
+                graphConfig: {
+                    data: {
+                        url: url.route('completedAssessmentsData', 'Tools', 'taoProctoring', { interval: config.interval })
```
ssipasseuth: It looks like every component still call its own data endpoint to refresh their data instead of being grouped as suggested.

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -191,94 +250,49 @@ define([
...
-                    renderTo : $container,
-                    actionName : msg,
-                    reason : true,
-                    allowedResources: [],
-                    reasonRequired: true,
-                    categoriesSelector: cascadingComboBox(categories['pause'])
-                };
+            // Activity Graph
+            updateActivityGraph({
+                autoRefresh: parseInt(activityGraphConfig.completed_assessments_auto_refresh, 10) * 1000,
```
ssipasseuth: still some hard coded interval to be replace with a server side configuration

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -196,7 +193,96 @@ define([
...
+        start: function () {
+            var autoRefreshInterval;
+            var config = $('.activity-dashboard').data('config');
+            var poll;
+
+            poll = polling({
+                action: function () {
+                    request(url.route('assessmentActivityData', 'Tools', 'taoProctoring', { interval: $activityGraphInterval.val() }))
+                    .then(function (data) {
+                        updateUserActivity(data.assessment_activity);
```
ssipasseuth: As commented in a previous review (sorry if it was not clear enough), each of the 4 "monitoring widgets" should be wrapped into their own components (extending the ui/component) and unit tested. 
Eventually, we should only have in the controller the initialization of the 4 components in the appropriate containers and their update after every polling interval.

config/default/ActivityMonitoringService.conf.php
```diff
@@ -10,4 +10,5 @@
 return new ActivityMonitoringService([
     ActivityMonitoringService::OPTION_ACTIVE_USER_THRESHOLD => 300,
     ActivityMonitoringService::OPTION_COMPLETED_ASSESSMENTS_AUTO_REFRESH => 30,
+    ActivityMonitoringService::AUTO_REFRESH_INTERVAL => 60,
```
hutnikau: Having these two properties may be confusing. I would suggest to combine them in one.

controller/Tools.php
```diff
@@ -45,36 +45,65 @@ class Tools extends SimplePageModule
...
+                'completed_assessments' => $completedAssessments
+            ]
+        ]);
     }
 
     /**
      * Get completed assessments data
      */
-    public function completedAssessmentsData()
+    private function completedAssessmentsData($param)
```
hutnikau: `$param` is not good name for parameter. 
Also I don't see the reason why you need to pass it. Controller has access to all the request parameters.

controller/Tools.php
```diff
@@ -112,20 +141,23 @@ public function pauseActiveExecutions()
...
+                'unprocessed' => $notPaused
+            ]
         ]);
     }
 
     /**
      * @return \DateInterval
      */
-    private function getInterval()
+    private function getInterval($param)
```
hutnikau: Retrieve interval value from request.

controller/Tools.php
```diff
@@ -143,21 +175,23 @@ private function getInterval()
...
         }
+
         return $interval;
     }
 
     /**
      * @return \DateTime[]
      */
-    private function getTimeKeys()
+    private function getTimeKeys($param)
```
hutnikau: Retrieve interval value from request

model/datatable/DeliveriesActivityDatatable.php
```diff
@@ -44,20 +44,22 @@ class DeliveriesActivityDatatable implements DatatablePayload, ServiceLocatorAwa
...
+        $this->request = DatatableRequest::fromGlobals();
     }
 
     public function getPayload()
     {
-        $service = $this->getServiceLocator()->get(ActivityMonitoringService::SERVICE_ID);
-        $data = $service->getData()['deliveries_statistics'];
+        if (is_null($this->data)) {
+            $service = $this->getServiceLocator()->get(ActivityMonitoringService::SERVICE_ID);
+            $data = $service->getData();
```
hutnikau: `$data` is not used. `DeliveriesActivityDatatable` will be broken if data is not passed to the constructor.

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -196,7 +193,96 @@ define([
...
+                    }
+                }
+            });
+        }
+    }
+
+    function updateDeliveryList(data) {
+        if (!$deliveryListDatatable) {
+            $deliveryListDatatable = $deliveryListContainer.datatable({
+                filter:                   false,
```
hutnikau: Add `url` option to load datatable data. Do not use direct injecting data into it. Otherwise ordering will be broken.

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -21,156 +21,153 @@
...
+        $('.authorized-but-not-started-assessments', $assessmentActivityContainer).text(data.authorized_but_not_started_assessments);
+
+        // Paused
+        $('.paused-assessments', $assessmentActivityContainer).text(data.paused_assessments);
+    }
+
+    function updateActivityGraph(data) {
+        if (!activityGraph) {
             activityGraph = activityGraphFactory({
-                autoRefresh : parseInt(activityGraphConfig.completed_assessments_auto_refresh, 10) * 1000,
```
hutnikau: Auto refresh progress bar is broken without `autoRefresh` config option.

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -21,156 +21,153 @@
...
 
-                bulkActionPopup(config).on('ok', function(reason){
-                    doPause(reason);
-                });
-            });
+    function updateAssessmentActivity(data) {
+        // Total Current Assessments
+        $('.total-current-assessments', $assessmentActivityContainer).text(data.total_current_assessments);
 
-            $('.js-activity-chart-interval').on('change', function () {
```
hutnikau: Select box does not work without it.

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -196,7 +193,96 @@ define([
                         show: false
                     }
                 }
-            }).render();
+            })
+            .render();
+        } else {
+            activityGraph.refresh({
```
hutnikau: after refreshing chart is broken:
image

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -21,156 +21,153 @@
...
-                                label : {
-                                    text:  interval === 'day' ? __('Hours') : __('Days')
-                                }
-                            }
-                        }
-                    }
-                });
-            });
+        // In Progress
+        $('.in-progress-assessments', $assessmentActivityContainer).text(data.in_progress_assessments);
```
hutnikau: Please consider to use handlebars template for this purpose instead of `.text()` calls.

ssipasseuth: It is indeed preferable to have all text update in updateAssessmentActivity() done via a handlebars template, in the assessmentActivity component

config/default/ActivityMonitoringService.conf.php
ssipasseuth: 👍 
Sorry, missed that one. 
Renaming it to OPTION_AUTO_REFRESH_INTERVAL is more consistent.

eacraft13:
I updated the pull request. I reverted back to using the completedAssessmentsData endpoint, as well as wrap the other pieces into widgets.

controller/Tools.php
eacraft13: For this one, I do feel it's best to retrieve the request parameter from the endpoint. It has better modularity. For instance, getInterval is now able to be unit tested without having to make an http request.

views/js/controller/Tools/assessmentActivity.js
eacraft13: I'm not aware of handlebars providing data binding. Can you provide an example of how to accomplish this?

views/js/component/activityMonitoring/currentAssessmentActivity/currentAssessmentActivity.tpl
```diff
@@ -0,0 +1,49 @@
+<div class="col-4 dashboard-block">
```
ssipasseuth: It may be better to wrap all the elements into a root div. This will allow to scope the whole component more efficiently, be it in css or javacript selector.

views/js/component/activityMonitoring/userActivity/userActivity.tpl
```diff
@@ -0,0 +1,18 @@
...
+</div>
+<div class="col-6 dashboard-block">
+    <span class="dashboard-block-number {{ activeProctors.container }}">
+        {{ activeProctors.value }}
+    </span>
+    <h3>
+        <span class="icon icon-test-taker"></span>
+        {{ activeProctors.label }}
+    </h3>
+</div>
```
ssipasseuth: same remark as currentAssessmentActivity.tpl about wrapping all element into a single div

views/js/component/activityMonitoring/deliveriesList/deliveriesList.js
```diff
@@ -0,0 +1,120 @@
...
+             * @param {Object} data - Data passed from taoProctoring/Tools/assessmentActivity endpoint
+             */
+            update: function update(data) {
+                this.$datatable.datatable('refresh', data);
+            }
+        }, _defaults)
+
+        .on('render', function () {
+            this.$datatable = this.getElement().datatable({
+                filter:                   false,
```
hutnikau: Ordering is still broken. You must set `url` option to allow datatable component load data from it.

views/js/controller/Tools/assessmentActivity.js
```diff
@@ -21,154 +21,87 @@
...
-                sortby : 'label'
-            }, deliveryListModel);
-            $deliveryList.datatable('refresh');
-
-
-            $('.js-pause').on('click', function() {
-                var config;
+            var assessmentActivityAutoRefreshInterval;
+            var completedAssessmentsAutoRefreshInterval;
+            var config = $('.activity-dashboard').data('config');
```
ssipasseuth: $('.activity-dashboard') is already defined as $container in line 77.

views/js/component/activityMonitoring/deliveriesList/deliveriesList.js
eacraft13: @aleh, I understand what you're referring to now. The problem is that we were trying to conserve backend calls. That is, previous changes to the pr were made to accommodate making one data call. I updated the backend to do the ordering. It definitely feels a bit hacky, but I don't believe updating ui/datatable is currently justified. Let me know your thoughts on this.

config/default/ActivityMonitoringService.conf.php
```diff
@@ -10,4 +10,5 @@
 return new ActivityMonitoringService([
     ActivityMonitoringService::OPTION_ACTIVE_USER_THRESHOLD => 300,
     ActivityMonitoringService::OPTION_COMPLETED_ASSESSMENTS_AUTO_REFRESH => 30,
+    ActivityMonitoringService::OPTION_ASSESSMENT_ACTIVITY_AUTO_REFRESH => 60,
```
eacraft13: I couldn't find a way to update a chart in c3 without a url that didn't cause an error that aleh previously pointed out. So, I reverted to using two data calls: one for assessment activity data, and the other for completed assessment data. I thought this ok because these two calls are fundamentally different. That is, assessment activity data retrieves it's data from the ActivityMonitoringService. And completed assessment data retrieves from the LoggerService. So, two auto refresh options are now used.

eacraft13: The filter should be a MM-DD format for all but 'last day' option. I'm not currently receiving this error. I'll have to look into why this is happening for you.

model/datatable/DeliveriesActivityDatatable.php
```diff
@@ -70,7 +72,7 @@ protected function doPostProcessing(array $result)
     {
         $payload = [
             'data' => $result,
-            'page' => (integer) $this->request->getPage(),
+            'page' => 1, // No pagination, so always page 1
```
hutnikau: Please do not do that. Pagination is supposed to be used there later

model/datatable/DeliveriesActivityDatatable.php
```diff
@@ -82,9 +84,9 @@ protected function doPostProcessing(array $result)
      */
     protected function doSorting(array &$result)
     {
-        $sortBy = $this->request->getSortBy();
-        $sortOrder = strcasecmp($this->request->getSortOrder(), 'asc') === 0 ? SORT_ASC : SORT_DESC;
-        $flag = ($sortBy === 'label') ? SORT_STRING | SORT_FLAG_CASE : SORT_NUMERIC;
+        $sortBy = 'label';
```
hutnikau: Sorting should be available for all the columns as it works now.

views/js/component/activityMonitoring/deliveriesList/deliveriesList.js
hutnikau: `url` option must be configured for datatable to be able to get data if user tries to sort data, switch to next/prev page, use filters. There is no another way to do that (you can't do it on client side). Obviously you need to have action in controller for this purpose. Also make sure that after auto update sorting will be preserved.

PR #366, opened 2017-05-08
Fix/tao 4308 proctor control error message details

eacraft13:
@hutnikau, if you have any suggestions for a better error message, please let me know.


hutnikau:
You handle only one case (if assessment has been finished) but there may be another cases.
Better solution will be use try-catch in `DeliveryHelper::terminateExecutions`. Something like that:
https://github.com/oat-sa/extension-tao-proctoring/compare/develop...fix/tao-4308_proctor_control_error_message_details__catch_exception?expand=1

eacraft13:
@hutnikau, I took some of your suggestions but not all.

- First, I updated the error message. I wanted it to be shorter and also give the user direction.

- Reviewing a little more, I understand your meaning about other cases and I updated the code to handle terminated tests as well.

- I also updated the setting and format of the processed and unprocessed data.

- I chose not to use exception handling to determine success of the delivery execution actions. I do think it a good way to flow through php code, as it is a common php convention I often see. But, at least from what I've seen, I don't think it's as prevalent here as conditionals or other error handling.

hutnikau:
Looks good! Thanks.

PR #392, opened 2017-06-14
fixed issue when Detailed Report doesn't work in IE

zagovorichev:


eacraft13:
Confirmed working on Windows 10 IE 11

PR #397, opened 2017-06-20
Feature/tao 4483 activate proctoring on selected deliveries

tikhanovichA:
Rollback changes in locales


scripts/update/Updater.php
```diff
@@ -350,5 +350,10 @@ public function update($initialVersion)
 
             $this->setVersion('5.12.0');
         }
+        
+        if ($this->isVersion('5.12.0')) {
```
boajer: Please consider `$this->skip('x', 'y');`

PR #398, opened 2017-06-20
Fix/tao 4514 fixed max attemps

tikhanovichA:


model/execution/StateService.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace oat\taoProctoring\model\execution;
+
+use oat\taoDelivery\model\execution\StateService as DeliverStateService;
+
+class StateService extends DeliverStateService
+{
+    public function getDeliveryStates()
```
antoinerobin: isn't it easier to add it to the ```oat\taoProctoring\model\implementation\DeliveryExecutionStateService``` class ?
Here you will loose every proctoring related behaviour

scripts/update/Updater.php
```diff
@@ -353,5 +354,12 @@ public function update($initialVersion)
 
         $this->skip('5.12.0', '5.12.1');
 
+        if ($this->isVersion('5.12.1')) {
+            $service = new StateService;
+            $service->setServiceManager($this->getServiceManager());
+            $this->getServiceManager()->register(StateService::SERVICE_ID, $service);
```
antoinerobin: if you add ```getDeliveryStates``` to the current proctoring state service you don't have to register a new service

model/implementation/DeliveryExecutionStateService.php
```diff
@@ -22,6 +22,7 @@
 
 use oat\oatbox\service\ServiceManager;
 use oat\taoDelivery\model\execution\DeliveryExecution;
+use oat\taoProctoring\model\execution\DeliveryExecution as ProctoringDeliveryExecution;
```
antoinerobin: this is already in the use statements

scripts/update/Updater.php
```diff
@@ -41,6 +41,7 @@
 use oat\taoProctoring\controller\Monitor;
 use oat\taoProctoring\controller\Tools;
 use oat\taoProctoring\model\authorization\AuthorizationGranted;
+use oat\taoProctoring\model\execution\StateService;
```
antoinerobin: this use is not needed anymore

PR #400, opened 2017-06-22
Added script for updating proctoring status for old deliveries

tikhanovichA:


scripts/UpdateOldDeliveriesProctoringStatus.php
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * @param array $params
+     * @return array
+     */
+    protected function getDeliveries($params = [])
+    {
+        if ($params) {
+            $deliveries = array();
```
Alroniks: And better move array definition to the beginning of the method to be sure that we'll always return an array.

PR #404, opened 2017-06-27
Fix/tao 4487 pause waiting deliveries

tikhanovichA:


scripts/TerminateNotStartedAssessment.php
```diff
@@ -86,10 +86,16 @@ public function __invoke($params)
                     $data[DeliveryMonitoringService::DELIVERY_EXECUTION_ID]
                 );
                 if ($testSessionService->isExpired($deliveryExecution)) {
-                    $deliveryExecutionStateService->cancelExecution($deliveryExecution, [
+                    $cancel = $deliveryExecutionStateService->cancelExecution($deliveryExecution, [
```
hutnikau: Better check if delivery execution is cancelable (`\oat\taoProctoring\model\implementation\DeliveryExecutionStateService::isCancelable`) and if so, cancel it or pause otherwise.

tikhanovichA: fixed

PR #411, opened 2017-07-03
added possibility to have many Proctor Services

zagovorichev:


model/ProctorServiceDelegator.php
```diff
@@ -0,0 +1,85 @@
...
+     * @var ProctorService
+     */
+    private $extendedService;
+
+    public function getResponsibleService()
+    {
+        if (!isset($this->extendedService))
+        {
+            foreach ($this->getOption(self::PROCTOR_SERVICE_HANDLERS) as $handler) {
+                if (!is_a($handler, ProctorService::class)) {
```
jbout: We should check for the interface, not the implementation

PR #412, opened 2017-07-03
Fix/tao 4576 unhelpful proctor warning message

eacraft13:
Bugfix - ACT
Jira: https://oat-sa.atlassian.net/browse/TAO-4576

Description:
As a proctor, when receiving an error upon terminating session, the message “No report available for this test session” is displayed, but that message does not make sense for the scenario (i.e., termination).

### DIFF
--- a/views/js/controller/Delivery/monitoring.js
+++ b/views/js/controller/Delivery/monitoring.js
@@ -454,37 +454,65 @@ define([
                  * @param {Object[]} deniedResources
                  */
                 function buildWarningMessage(action, selection, deniedResources) {
+                    var maxReasons = 2;
                     var warningAction;
-                    var warningReasons;
-                    var warningMessage;
+                    var warningReason;
 
-                    if (deniedResources) {
-                        if (deniedResources.length > 1) {
-                            warningAction = __('Cannot %s these test sessions.', action);
-                            warningReasons = _(deniedResources)
-                                .map(function (deniedResource) {
-                                    return __('Test %s %s.', deniedResource.id, deniedResource.reason);
-                                })
-                                .slice(0, 2)
-                                .join(' ');
+                    deniedResources = deniedResources || [];
+                    selection = selection || [];
 
-                            if (deniedResources.length > 2) {
-                                warningReasons += __('...');
-                            }
+                    warningAction = getWarningAction(action, deniedResources.length > 1 || selection.length > 1);
+                    warningReason = _(deniedResources || [{}])
+                        .map(function (deniedResource, index) {
+                            return getWarningReason(deniedResource.reason, deniedResource.id || (index + 1));
+                        })
+                        .slice(0, maxReasons)
+                        .join(' ');
 
-                            warningMessage = __('%s %s', warningAction, warningReasons);
+                    if (deniedResources.length > maxReasons) {
+                        warningReason += __('...');
+                    }
+
+                    return warningAction + ' ' + warningReason;
+
+                    function getWarningAction(actionName, isPlural) {
+                        if (isPlural) {
+                            return {
+                                authorize: __('Cannot authorize test sessions.'),
+                                pause: __('Cannot pause test sessions.'),
+                                print: __('Cannot print test sessions.'),
+                                report: __('Cannot generate report on test sessions.'),
+                                terminate: __('Cannot terminate test sessions.'),
+                                time: __('Cannot extend test sessions.'),
+                                default: __('No report available for test sessions.')
+                            }[actionName || 'default'];
                         } else {
-                            warningMessage = __('Cannot %s this test session because test %s.', action, deniedResources[0].reason);
-                        }
-                    } else {
-                        if (selection.length > 1) {
-                            warningMessage = __('No report available for these test sessions.');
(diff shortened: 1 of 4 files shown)


views/js/controller/Delivery/monitoring.js
```diff
@@ -434,13 +434,33 @@ define([
...
+                    var warningAction;
+                    var warningMessage;
+                    var warningReasons;
 
                     if (!config.allowedResources.length) {
-                        if (_selection.length > 1) {
-                            feedback().warning(__('No report available for these test sessions'));
+                        if (config.deniedResources) {
+                            if (config.deniedResources.length > 1) {
+                                warningAction = __('Cannot') + ' ' + actionName + ' ' + __('these test sessions.');
```
jsconan: This is not well translatable. English may be correct, not the other languages. You should rather use a map of labels, like this:
```javascript
var warningMessages =  {
    pause: __('Cannot pause these test sessions.'),
    terminate: __('Cannot terminate these test sessions.'),
    ...
};
warningAction = warningMessages[actionName];
```

views/js/controller/Delivery/monitoring.js
```diff
@@ -434,13 +434,33 @@ define([
...
+                                warningReasons += _.map(config.deniedResources, function (deniedResource, index) {
+                                    var opener = (index > 0 ? __('test') : __('Test')) + ' ';
+                                    return opener + deniedResource.id + ' ' + deniedResource.reason;
```
jsconan: The translate helper accept parametric sentences, so you should rather use this construction instead: `_('Test %s', deniedResource.id)`.

And if you have to adapt the sentence to the context, then you will have to call the translate helper for each case. The reason why is we use a scanner to gather all translations from source code, so the labels should be only simple atomic strings.

views/js/controller/Delivery/monitoring.js
```diff
@@ -434,13 +434,33 @@ define([
...
+                                }).join(',') + '.';
+                                warningMessage = warningAction + ' ' + warningReasons;
+                            } else {
+                                warningAction = __('Cannot') + ' ' + actionName;
+                                warningReasons = __('this test session') + ' ' + config.deniedResources[0].reason;
+                                warningMessage = warningAction + ' ' + __('because') + ' ' + warningReasons;
```
jsconan: Please do not cut the sentences that should be translated. Use parametric sentences instead. If needed you may also use a map of sentences, like described earlier.

When building a message that should be translated, always thing about the translation it could have.

eacraft13:
Sorry to forget the ticket reference. But thanks for the heads up to 'saved replys'.

views/js/controller/Delivery/monitoring.js
```diff
@@ -452,6 +448,78 @@ define([
...
+
+                    warningAction = getWarningAction(action, deniedResources.length > 1 || selection.length > 1);
+                    warningReason = _(deniedResources || [{}])
+                        .slice(0, maxReasons)
+                        .map(function (deniedResource, index) {
+                            return getWarningReason(deniedResource.reason, deniedResource.id || (index + 1));
+                        })
+                        .join(' ');
+
+                    return __('%s %s%s', warningAction, warningReason, (deniedResources.length > maxReasons ? warningReasonOmission : ''));
```
jsconan: I think there is no need to use the translate helper to build such a string: there is nothing to translate, just a few placeholders. So you should just concat the strings.

views/js/helper/status.js
```diff
@@ -67,21 +67,21 @@ define(['lodash', 'i18n'], function(_, __){
             code : _canceled,
             label : __('Canceled'),
             can : {
-                authorize : __('is canceled'),
-                pause : __('is canceled'),
-                terminate : __('is canceled'),
+                authorize : __('canceled'),
+                pause : __('canceled'),
+                terminate : __('canceled'),
```
jsconan: It seems you are using these labels as a key to get the translated message. If so you should not translate them, otherwise you will go in trouble if those labels are translated to something else...

views/js/controller/Delivery/monitoring.js
```diff
@@ -452,6 +448,78 @@ define([
...
+                                print: __('Cannot print test session.'),
+                                report: __('Cannot generate report on test session.'),
+                                terminate: __('Cannot terminate test session.'),
+                                time: __('Cannot extend test session.'),
+                                default: __('No report available for test session.')
+                            }[actionName || 'default'];
+                        }
+                    }
+
+                    function getWarningReason(reason, testId) {
```
jsconan: Those two function are nice, but maybe should they be placed into a proper helper, like `taoProctoring/helper/status`, should not be ?

views/js/controller/Delivery/monitoring.js
jsconan: FYI, if you need a format string, you could also rely on the module `core/format`. This is is the same helper used by the translate helper to format parametric translations.
So here you'd just need to replace `__` by `format`, after having required the proper module of course.

eacraft13:
@jsconan I've refactored to address your concerns and move the creation of the warning message to the status helper.

PR #415, opened 2017-07-06
Fix/tao 4528 assessment activity screen

tikhanovichA:
https://oat-sa.atlassian.net/browse/TAO-4528


model/ActivityMonitoringService.php
```diff
@@ -58,6 +85,13 @@ class ActivityMonitoringService extends ConfigurableService
         DeliveryExecution::STATE_FINISHIED,
     ];
 
+    protected $assessmentsDeliveryStatusesMaps = [
```
eacraft13: I wonder if this would be better placed in `getRetiredDeliveries` function. This is opinion, so if you disagree that is acceptable. My argument would be that this is a one off property used in one place for one purpose.

PR #422, opened 2017-07-16
Feature/tao 4733 lti extended time

tikhanovichA:


model/execution/DeliveryExecutionManagerService.php
```diff
@@ -133,11 +156,13 @@ public function setExtraTime($deliveryExecutions, $extraTime = null)
 
             /** @var QtiTimer $timer */
             $timer = $this->getDeliveryTimer($deliveryExecution);
-            $timer->setExtraTime($extraTime)->save();
+            $timer->setExtraTime($extraTime);
+            $timer->setExtendedTime($extendedTime)->save();
```
boajer: Why not to use `$timer->setExtraTime($extraTime)->setExtendedTime($extendedTime)->save();` ?

model/execution/DeliveryExecutionManagerService.php
```diff
@@ -80,21 +81,43 @@ public function getDeliveryTimer($deliveryExecution)
      * Sets the extra time to a list of delivery executions
      * @param $deliveryExecutions
      * @param null $extraTime
+     * @param null $extendedTime
      * @return array
      */
-    public function setExtraTime($deliveryExecutions, $extraTime = null)
+    public function setExtraTime($deliveryExecutions, $extraTime = null, $extendedTime = null)
```
boajer: Could you update also `\oat\taoProctoring\helpers\DeliveryHelper::setExtraTime` signature

model/execution/DeliveryExecutionManagerService.php
```diff
@@ -80,21 +81,43 @@ public function getDeliveryTimer($deliveryExecution)
...
+                $inputParameters = $testSessionService->getRuntimeInputParameters($deliveryExecution);
+                /** @var AssessmentTest $testDefinition */
+                $testDefinition = \taoQtiTest_helpers_Utils::getTestDefinition($inputParameters['QtiTestCompilation']);
+                $deliveryExecutionArray[] = $deliveryExecution;
+                $extraTime = null;
+                if ($maxTime = $testDefinition->getTimeLimits()->getMaxTime()) {
+                    $seconds = $maxTime->getSeconds(true);
+                    $secondsNew = $seconds * $extendedTime;
+                    $extraTime = floor(($secondsNew - $seconds) / 60) * 60;
+                    $deliveryMonitoringService = $this->getServiceLocator()->get(DeliveryMonitoringService::SERVICE_ID);
```
boajer: `deliveryMonitoringService` is already in this scope, I think there is no reason to redefine it

model/monitorCache/implementation/DeliveryMonitoringData.php
```diff
@@ -255,7 +256,14 @@ private function updateExtraTime()
             $timer->setStorage(new QtiTimeStorage($this->deliveryExecution->getIdentifier(), $this->deliveryExecution->getUserIdentifier()));
             $timer->load();
         }
-        $this->addValue(DeliveryMonitoringService::EXTRA_TIME, $timer->getExtraTime(), true);
+        if ($testPart = $testSession->getCurrentTestPart()) {
+            if ($testSessionLimits = $testPart->getTimeLimits()) {
+                $maxTime = $testSessionLimits->getMaxTime()->getSeconds(true);
```
jsconan: Caution, `$testSessionLimits->getMaxTime()` could be null

manifest.php
```diff
@@ -44,7 +44,7 @@
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '6.0.0',
+    'version' => '6.1.0',
```
jsconan: As there is an API dependency to taoQtiTest you should also update the required version, below. Otherwise the runtime could fail as the API related to the extended time could be missing.

views/js/controller/Delivery/monitoring.js
```diff
@@ -873,6 +873,15 @@ define([
                         });
                     }
 
+                    model.push({
+                        id: 'extendedTime',
+                        label: __('Extended Time'),
+                        transform: function(value, row) {
+                            var timer = _.isObject(row.timer) ? row.timer : {};
+                            return timer.extendedTime;
```
jsconan: As the extended time is a factor, maybe should it be better to illustrate it. For instance by prefixing the displayed value with a `x`. This way it could be: `x1.5`. What do you think?

tikhanovichA: good idea. I'll add it. Thanks

PR #441, opened 2017-08-02
Fix user names in irregularity report

hutnikau:


model/service/IrregularityReport.php
```diff
@@ -75,4 +77,23 @@ public function getIrregularitiesTable(\core_kernel_classes_Resource $delivery,
...
     }
 
+    /**
+     * @param \core_kernel_classes_Resource $user
+     * @return string
+     */
+    private function getUserName(\core_kernel_classes_Resource $user)
+    {
+        if (!isset($this->userNames[$user->getUri()])) {
+            $userName = $user->getLabel();
```
boajer: May be `\oat\tao\helpers\UserHelper::getUserName` should be reused here?

boajer: What about `\oat\tao\helpers\UserHelper::getUserName` ?

PR #446, opened 2017-08-14
Fix/tao 4922 extended time

tikhanovichA:


model/execution/DeliveryExecutionManagerService.php
```diff
@@ -106,12 +109,44 @@ public function setExtraTime($deliveryExecutions, $extraTime = null, $extendedTi
...
+
+                    if ($section = $testSession->getCurrentAssessmentSection()) {
+                        if ($testSessionLimits = $section->getTimeLimits()) {
+                            $seconds = $testSessionLimits->hasMaxTime()
+                                ? $testSessionLimits->getMaxTime()->getSeconds(true)
+                                : $seconds;
+                        }
+                    }
+
+                    if ($item = $testSession->getCurrentAssessmentItemRef()) {
```
hutnikau: It's better to check items timers first and then if `$seconds` is `null` search it in upper elements

PR #452, opened 2017-08-29
Fix/tao 4968/session history 500 error

chilly86:


eacraft13:
Your fix introduces another bug. Now that you can get to session history, returning to session will produce another 500 error (although it is still able to get back to that page in this case). Would you be able to provide this fix here as well?

PR #465, opened 2017-09-14
Fix/assessment activity

zagovorichev:


hutnikau:
Direct query to the sql storage is not an option. We have delivery monitoring service which does the same query.
Moreover you do not take into account deliveries which was not launched yet because there will no records in `delivery_monitoring` table for such deliveries.
This is alternative solution which gives some performance boost:
https://github.com/oat-sa/extension-tao-proctoring/pull/466

zagovorichev:
@hutnikau This is the correct option. Your fix doesn't work for the real data, because when you're getting all data from the delivery monitoring table then you trying to go through all of them and it takes too many time. (for example you could try delivery monitoring with 100 000 records).
I didn't knew that we need all deliveries but it's easy to fix.

hutnikau:
@zagovorichev 
1. you can't do direct query to rds storage. In this case we will not be able to change delivery monitoring implementation.
2. you code returns wrong results. This implementation does not take into account deliveries without executions.

zagovorichev:
1. I can move it to the delivery monitoring service
2. Quick fix for that I could take from your previous code
```
$deliveries = DeliveryAssemblyService::singleton()->getAllAssemblies();
foreach ($deliveries as $delivery) {
      $newResult[$delivery->getUri()] = $statusesArray;
      $newResult[$delivery->getUri()]['label'] = $delivery->getLabel();
}
```
3. Your decision doesn't work

Would that suit you?

hutnikau:
If you will add this code then it will not solve the problem. We again will have to go through all thousands of deliveries on `foreach` loop

zagovorichev:
1. You're going through delivery executions, not a delivery (worse then delivery with counted executions)
2. In my code we could implement pagination on the RDS side

hutnikau:
You are right. But as I said you can't make query to storage from there. Add it to delivery monitoring service. Do not forget to change interface and it would be great to have unit test for this method.

zagovorichev:
@hutnikau Update

model/ActivityMonitoringService.php
```diff
@@ -237,31 +234,75 @@ protected function getNumberOfActiveUsers($role = null)
...
+                        if ($status == 'label') {
+                            continue;
+                        }
+                        $list[self::FIELD_RETIRED_DELIVERIES][$status] = (int)$list[self::FIELD_RETIRED_DELIVERIES][$status] + (int)$val;
+                    }
+                }
+            }
+        }
+
+        return ['data' => $list, 'total' => $total];
```
hutnikau: Currently this function is supposed to be used with datatable. You have to return only `$list` variable and wrap it into `['data' => ..., $total => ...` array inside `DeliveriesActivityDatatable`

model/ActivityMonitoringService.php
```diff
@@ -237,31 +234,75 @@ protected function getNumberOfActiveUsers($role = null)
...
-        $newResult[self::FIELD_RETIRED_DELIVERIES] = $statusesArray;
-        $newResult[self::FIELD_RETIRED_DELIVERIES]['label'] = self::LABEL_RETIRED_DELIVERIES;
+        /** @var DeliveryMonitoringService $deliveryMonitoringService */
+        $deliveryMonitoringService = $this->getServiceManager()->get(DeliveryMonitoringService::SERVICE_ID);
+        $deliveryService = DeliveryAssemblyService::singleton();
 
-        $deliveries = DeliveryAssemblyService::singleton()->getAllAssemblies();
+        $limit = isset($params['rows']) ? $params['rows'] : 10;
+        $offset = isset($params['page']) ? ($params['page']-1) * $limit : 0;
+        $deliveries = $deliveryService->getRootClass()->getInstances(true, [
```
hutnikau: Pagination is not responsibility of `ActivityMonitoringService`. It's responsibility of `DeliveriesActivityDatatable` class. You have to get array of deliveries in `DeliveriesActivityDatatable` (according to `rows` and `page` values) and pass this array into `ActivityMonitoringService::getStatesByDelivery()`.

model/ActivityMonitoringService.php
```diff
@@ -237,31 +234,75 @@ protected function getNumberOfActiveUsers($role = null)
...
+            'order' => RDFS_LABEL,
+            'offset' => $offset,
+            'limit' => $limit
+        ]);
+
+        // deliveries count + retired deliveries row
+        $total = $deliveryService->getRootClass()->countInstances([], ['recursive' => true]) + 1;
```
hutnikau: Please count total amount of deliveries inside of `DeliveriesActivityDatatable` class.

zagovorichev:
@hutnikau Update

model/ActivityMonitoringService.php
```diff
@@ -131,8 +129,6 @@ public function getData()
             self::STATE_IN_PROGRESS_ASSESSMENTS => $active
         ];
 
-        $deliveryStates = $this->getStatesByDelivery();
```
zagovorichev: Also I deleted this, because I didn't find where this used, please, check it twice

model/datatable/DeliveriesActivityDatatable.php
```diff
@@ -42,19 +43,39 @@ class DeliveriesActivityDatatable implements DatatablePayload, ServiceLocatorAwa
...
+            'sort_by' => $this->request->getSortBy(),
+            'sort_order' => $this->request->getSortOrder(),
+            'filters' => $this->request->getFilters()];
+
+        $limit = isset($params['rows']) ? $params['rows'] : 10;
+        $offset = isset($params['page']) ? ($params['page']-1) * $limit : 0;
+        $deliveries = $this->deliveryService->getRootClass()->getInstances(true, [
+            'order' => RDFS_LABEL,
+            'offset' => $offset,
+            'limit' => $limit
```
hutnikau: You can't do pagination in this way. You always have to iterate through all existing deliveries. Otherwise ordering will work only for the current page.

zagovorichev: yes, I know, but it task become to complex with that improvement, so for now it will be ordering only on the current page but it will work on prod

hutnikau: In this case remove pagination from this datatable at all. We can't add feature which does not work properly.

zagovorichev: We need this pagination because with 2000 deliveries it won't work too (it's a part of my fix)

zagovorichev: I could remove your sorting. I need to make quick fix for your code but not to create new feature

hutnikau: Sorting by all columns was requested by ACT. Your PR is breaking this feature.

zagovorichev:
@hutnikau Update

PR #475, opened 2017-10-11
Allow configurable persistance on installation stage

boajer:


scripts/install/SetupDeliveryMonitoring.php
```diff
@@ -37,13 +36,21 @@ class SetupDeliveryMonitoring extends InstallAction
      */
     public function __invoke($params)
     {
-        $pm = $this->getServiceLocator()->get(\common_persistence_Manager::SERVICE_ID);
-        $persistence = $pm->getPersistenceById('default');
+        try {
+            $service = $this->getServiceManager()->get(MonitorCacheService::SERVICE_ID);
+        } catch (\Exception $exception) {
```
hutnikau: it's better to catch `\oat\oatbox\service\ServiceNotFoundException`

scripts/install/SetupDeliveryMonitoring.php
```diff
@@ -37,13 +37,21 @@ class SetupDeliveryMonitoring extends InstallAction
...
+        }
+        $persistence = $service->getPersistence();
+
         DbSetup::generateTable($persistence);
-        $this->registerService(DeliveryMonitoringService::SERVICE_ID, new MonitorCacheService(array(
-            MonitorCacheService::OPTION_PERSISTENCE => 'default',
-            MonitorCacheService::OPTION_PRIMARY_COLUMNS => DbSetup::getPrimaryColumns()
-        )));
+
+        $service->setOption(MonitorCacheService::OPTION_PRIMARY_COLUMNS, DbSetup::getPrimaryColumns());
```
hutnikau: this option already set in the constructor

boajer: fixed

PR #482, opened 2017-10-20
remove debug comments

krampstudio:
It looks like I've left debug comments in the [previous PR](https://github.com/oat-sa/extension-tao-proctoring/pull/480)

### DIFF
--- a/model/execution/ProctoredSectionPauseService.php
+++ b/model/execution/ProctoredSectionPauseService.php
@@ -52,7 +52,6 @@ class ProctoredSectionPauseService extends SectionPauseService
             /** @var AssessmentItemRef $itemRef */
             $itemRef = $session->getCurrentAssessmentItemRef();
 
-            \common_Logger::i(' > CURRENT item ref is ' . $itemRef->getIdentifier());
             return $this->isProctored($session) && $this->isItemPausable($itemRef);
         }
         return false;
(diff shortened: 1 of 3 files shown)


model/execution/ProctoredSectionPauseService.php
```diff
@@ -68,17 +67,11 @@ public function isPausable(TestSession $session = null)
...
 
                 if ($this->isItemPausable($itemRef)) {
-                    \common_Logger::i(' > item pauseable');
                     return false;
-                } else {
-                    
-                    \common_Logger::i(' > PREVIOUS item not pauseable');
                 }
         }
         return true;
```
zagovorichev:
```
return ! ($session->getState() === AssessmentTestSessionState::INTERACTING
    && $this->isProctored($session)
    && $this->isItemPausable($session->getCurrentAssessmentItemRef()))
```

model/execution/ProctoredSectionPauseService.php
```diff
@@ -116,19 +109,8 @@ private function isProctored(TestSession $session)
...
-                \common_Logger::i(' >  ITEM REF not  category ' . $itemRef->getIdentifier() . ' -> ' . self::PAUSE_CATEGORY);
-
-            } 
-            return $pauseable;
+            return $itemRef->getCategories()->contains(self::PAUSE_CATEGORY);
         }
-             else {
-                \common_Logger::i(' >  ITEM REF IS NULL');
-            }
         return false;
```
zagovorichev: only one exit point could be used
```
return is_null($itemRef) ? false : $itemRef->getCategories()->contains(self::PAUSE_CATEGORY);
```

krampstudio:
I've simplified one of the condition however I like to keep the default return value as explicit as possible

PR #486, opened 2017-10-25
Fix/tao 5339/display message on the receiver side

ionutpad:


chilly86:
- it requires your new version of taoQtiTest

PR #492, opened 2017-10-30
Fix/tao 5361 persist extra time

jsconan:
Related to: 
- https://oat-sa.atlassian.net/browse/TAO-5361
- https://oat-sa.atlassian.net/browse/TAO-5313

Since the refactoring of the ExtendedState, we must explicitly persist the data that should go to the state storage. The extra time feature has been ignored by this refactoring.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -44,7 +44,7 @@ return array(
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '7.8.3',
+    'version' => '7.8.4',
     'author' => 'Open Assessment Technologies SA',
     'requires' => array(
         'tao' => '>=10.28.0',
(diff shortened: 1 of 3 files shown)


manifest.php
```diff
@@ -44,7 +44,7 @@
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '7.8.3',
+    'version' => '7.8.4',
```
krampstudio: The `StorageManager` has been introduced on a recent version of `taoQtiTest` (`~17.12.0`). So please increase both feature number and the dependency to taoQtiTest

krampstudio:
How to test this ? Just ensure extra tiem works correctly ?

manifest.php
jsconan: I would say the version has already been upgraded, but it seems there was a merge issue and the dependency was not updated when the conflict on the main PR has been resolved.
So I will fix that.

jsconan:
Yes, to test just ensure the extra time is working properly. With the refactoring of Extended State, the extra time was not properly recorded, so:
- just after the proctor has granted the extra time, if you check the value it should be reset to 0
- the test taker never see any additional time granted

With the patch, both issues should have been fixed

PR #509, opened 2017-11-27
Tao 5558 fix default sorting

hutnikau:


model/implementation/TestSessionHistoryService.php
```diff
@@ -280,7 +280,11 @@ private function sortHistory(array &$history, array $options)
         }
         if ($sortBy == 'timestamp' || $sortBy == 'id') {
             usort($history, function($a, $b) use($sortOrder) {
-                return $sortOrder * ($a['timestamp'] - $b['timestamp']);
```
hutnikau: Example:
`1511510824.289 - 1511510824.2889` will return `-0.0001` but as per documentation value should be an integer:
http://php.net/manual/en/function.usort.php

PR #519, opened 2017-12-11
Refactoring/tao 5104/remove globals from generis

ionutpad:
Depends on https://github.com/oat-sa/extension-tao-delivery-rdf/pull/156


helpers/DeliveryHelper.php
```diff
@@ -107,12 +108,12 @@ public static function buildDeliveryData($delivery, $executions)
         }
         
         $deliveryProps = array(
-            new \core_kernel_classes_Property(TAO_DELIVERY_START_PROP),
-            new \core_kernel_classes_Property(TAO_DELIVERY_END_PROP),
+            new \core_kernel_classes_Property(DeliveryContainerService::PROPERTY_START),
```
zagovorichev: I think that this wrong place for `PROPERTY_START`, it should be here `DeliveryAssemblyService`

zagovorichev:
@jbout Could you check that constants have correct site location, please?

helpers/DeliveryHelper.php
ionutpad: The constant defined in the DeliveryAssemblyService are constants related to Assembled Delivery based on this fact I have separate them.

zagovorichev: properties of the delivery should be in the one place
confirmed by @jbout

jbout: these are properties related to the assembled delivery, they describe the time window during which the tt is allowed to start the test

zagovorichev:
waiting for https://github.com/oat-sa/extension-tao-delivery-rdf/pull/156

PR #526, opened 2017-12-20
TAO-5705 - NCCER - Close Abandoned Script

bugalot:


model/implementation/DeliveryExecutionStateService.php
```diff
@@ -216,6 +213,12 @@ public function terminateExecution(DeliveryExecution $deliveryExecution, $reason
                 $this->getTestSessionService()->persist($session);
                 $this->getServiceLocator()->get(ExtendedStateService::SERVICE_ID)->persist($session->getSessionId());
             }
+            
+            // Delivery execution state changes after test session ends, in the same way as it happens
+            // when a human test taker takes the test.
+            $eventManager->trigger(new DeliveryExecutionTerminated($deliveryExecution, $proctor, $reason));
```
bugalot: @jsconan  **Be very careful here ! I changed the order of the events being triggered.** @antoinerobin and @bugalot do not think that's a problem, but it's worth to triple check it does not break anything elsewhere.

jsconan: Should not be a problem. I checked the use of this event and the order does not matter.

scripts/TerminatePausedAssessment.php
```diff
@@ -131,4 +131,18 @@ protected function addReport($type, $message)
...
+     * 
+     * Provides the 'reasons' information array with keys 'category' and 'subCategory'.
+     * This method may be overriden by subclasses to provide customer specific information.
+     * 
+     * @return arrray.
+     */
+    protected function getTerminationReasons()
+    {
+        // @fixme remove customer specific information.
+        return ['category' => 'Technical', 'subCategory' => 'ACT'];
```
jsconan: Not related to this story, but it would better to set this content in the proper extension. I don't think the `subCategory` is actually generic :)

PR #537, opened 2018-01-18
Tao 5822 move time keys to date helper

hutnikau:
depends on https://github.com/oat-sa/tao-core/pull/1617

### DIFF
--- a/model/ActivityMonitoringService.php
+++ b/model/ActivityMonitoringService.php
@@ -141,6 +141,7 @@ class ActivityMonitoringService extends ConfigurableService
 
     /**
      * @see \tao_helpers_Date::getTimeKeys()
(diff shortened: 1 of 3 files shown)


model/ActivityMonitoringService.php
```diff
@@ -140,69 +140,12 @@ public function getData()
...
-     *       'timezone' => '+00:00',
-     *     )),
-     *       ...
-     *   )
-     *
-     * @param \DateInterval $interval
-     * @param \DateTime|null $date
-     * @param null $amount
-     * @return \DateTime[]
+     * @see \tao_helpers_Date::getTimeKeys()
```
Alroniks: Seems that @deprecated annotation is missed.

PR #545, opened 2018-02-12
Fix/tao 5282/proctor timer

zagovorichev:


views/js/controller/Delivery/monitoring.js
```diff
@@ -1055,6 +1048,11 @@ define([
                                 });
                             }
                             loadingBar.stop();
+
+                            appController.on('change', function () {
```
zagovorichev: @jsconan I've got new issue, that I can't to do `off('chage')` and now each polling.stop will be accumulated. Maybe we have something like `once` to delete it after it worked

zagovorichev: @krampstudio ? :)

zagovorichev: thank you!

zagovorichev:
@krampstudio Update

PR #550, opened 2018-02-25
Tao 5970/searchable columns

boajer:
Allows first and last name to be filtered by in proctoring
Introduce cli tool to migrate proctoring monitoring data from KV to RDBS
https://oat-sa.atlassian.net/browse/TAO-5970


scripts/tools/KvMonitoringMigration.php
```diff
@@ -0,0 +1,193 @@
...
+                'description' => 'Indicate whether or not changes in config should be recorded. Please note that applies to the current instance only and changes MUST be distributed across all of them'
+            ],
+
+            'chunkSize' => [
+                'prefix' => 'c',
+                'defaultValue' => 100,
+                'longPrefix' => 'chunk-size',
+                'cast' => 'integer',
+                'required' => false,
+                'description' => 'The cache ttl for the delivery theme provider'
```
hutnikau: Is that correct description?

PR #563, opened 2018-03-14
Feature/tao 6088 resend completedtestevent irregularity report delivery finished

bugalot:


model/event/DeliveryExecutionIrregularityReport.php
```diff
@@ -0,0 +1,54 @@
...
+class DeliveryExecutionIrregularityReport implements Event
+{
+    private $deliveryExecution;
+
+    /**
+     * @return string
+     */
+    public function getName()
+    {
+        return __CLASS__;
```
hutnikau: I would suggest to use `static::class` here because this is not a final class and if there will be one more class which extends this one `getName` function will return name of parent class (where `__CLASS__` constant has been declared).

PR #574, opened 2018-04-25
Feature/tao 5531 make report url configurable

siwane:


model/AssessmentResultsService.php
```diff
@@ -187,6 +189,20 @@ public function getPrintableRubric(DeliveryExecutionInterface $deliveryExecution
...
 
+    /**
+     * Get the configurable url parts or default url to proctoring
+     */
+    public function getScoreReportUrlParts()
+    {
+        return $this->hasOption(self::OPTION_SCORE_URL) && is_array($this->getOption(self::OPTION_SCORE_URL))
+            ? $this->getOption(self::OPTION_SCORE_URL)
+            : array(
+                'extension' => 'taoProctoring',
```
ionutpad: I would prefer having this kind of config setup by default in in taoProctoring/config/default/AssessmentResults.conf.php and overwrite by nccer-delivery, it would make things easier to detect the default config.

scripts/update/Updater.php
```diff
@@ -636,6 +636,6 @@ public function update($initialVersion)
             $this->setVersion('8.7.0');
         }
 
-        $this->skip('8.7.0', '8.7.1');
+        $this->skip('8.7.0', '8.8.0');
```
ionutpad: The updating will require also updating the config with:
```php
'score_report_url' => array(
        'extension' => 'taoNccerDelivery',
        'controller' => 'ScoreReport',
        'action' => 'printing'
    )
```
Because in case it's not set we are returning this anyway.

views/js/controller/Delivery/monitoring.js
```diff
@@ -371,14 +367,17 @@ define([
 
                 // print the score reports
                 function printReport(selection) {
-                    print(selection, 'printReport');
+                    execBulkAction('print', __('Print Score'), selection, function(sel) {
+                        var url = urlHelper.route(
+                            printReportUrl.action ||'printReport',
+                            printReportUrl.controller ||'Reporting',
+                            printReportUrl.extension ||'taoProctoring',
```
ionutpad: Having default config this should disappear as well.

PR #584, opened 2018-05-14
TAO-6279

ionutpad:


model/monitorCache/implementation/MonitoringStorage.php
```diff
@@ -803,13 +805,16 @@ public function getStatusesStatistic($limit = 10, $offset = 0, $orderby = 'label
...
         foreach ($statusesMap as $label => $statusUri) {
             $queryBuilder->leftJoin(
                 'delivery_m',
                 self::TABLE_NAME,
                 $conn->quoteIdentifier('s_'.$label),
-                'delivery_m.delivery_execution_id='.$conn->quoteIdentifier('s_'.$label).'.delivery_execution_id and '.$conn->quoteIdentifier('s_'.$label).'.status = \''.$statusUri.'\''
+                'delivery_m.delivery_execution_id='.$conn->quoteIdentifier('s_'.$label).'.delivery_execution_id and '
+                .$conn->quoteIdentifier('s_'.$label).'.status = :status_uri'
             );
+            $paramsValues[':status_uri'] = $statusUri;
```
hutnikau: you override ':status_uri' parameter on each iteration

ionutpad: Good catch, thank you for the fix.

PR #600, opened 2018-06-07
Hotfix/tao 6278/delete exec cancelled

ionutpad:


manifest.php
```diff
@@ -89,6 +90,8 @@
             \oat\taoProctoring\scripts\install\RegisterProctoringRunnerService::class,
             \oat\taoProctoring\scripts\install\SetupProctorCsvImporter::class,
             \oat\taoProctoring\scripts\install\RegisterProctorAttemptService::class,
+            \oat\taoProctoring\scripts\install\RegisterProctoringRunnerService::class,
```
hutnikau: looks like this like is redundant

manifest.php
```diff
@@ -45,12 +46,12 @@
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '8.13.3',
+    'version' => '8.13.4',
```
hutnikau: please increase minor version as you created a new service

PR #616, opened 2018-07-23
Feature/tao 6521/sync irregularities

ionutpad:


model/deliveryLog/implementation/RdsDeliveryLogService.php
```diff
@@ -152,12 +153,13 @@ public function insertMultiple(array $data)
      *      'limit' => null, // to get all records
      *      'offset' => 0,
      *  ]
+     * @param bool $shouldDecodeData
      * @return mixed
      */
-    public function search($params = [], $options = [])
+    public function search($params = [], $options = [], $shouldDecodeData = true)
```
siwane: Could u pass `shouldDecodeData` in the options parameter please. In this case you do not need to change the interface

PR #628, opened 2018-09-20
Tao 6876 improve cache

alexeyvy:
According to https://oat-sa.atlassian.net/browse/TAO-6876


model/monitorCache/implementation/DeliveryMonitoringData.php
```diff
@@ -77,16 +78,23 @@ class DeliveryMonitoringData implements DeliveryMonitoringDataInterface, Service
...
         if (is_array($data) && !empty($data)) {
             $this->data = $data;
         } else {
-            $this->data = [DeliveryMonitoringService::DELIVERY_EXECUTION_ID => $deliveryExecution->getIdentifier()];
+            $this->data = [
+                DeliveryMonitoringService::DELIVERY_EXECUTION_ID => $deliveryExecution->getIdentifier(),
+            ];
+        }
+
+        if (!array_key_exists(DeliveryMonitoringService::STATUS, $this->data) && $deliveryExecution->getState()) {
```
hutnikau: Why we need set execution state in constructor?
Even if we need it I believe we can completely remove `if` condition.

model/monitorCache/implementation/MonitoringStorage.php
```diff
@@ -328,27 +320,60 @@ public function count(array $criteria = [])
...
+                // and not the number of rows matched by the where clause.
+                // So just in case try to create without fallback
+                try {
+                    $this->create($deliveryMonitoring);
+                } catch (\PDOException $e) {
+                } catch (DBALException $e) {
+                }
+            }
+            $this->saveKvData($deliveryMonitoring);
+            $result = true;
```
hutnikau: `$result` will be `true` even in case of `PDOException` and `DBALException`.

helpers/DeliveryHelper.php
```diff
@@ -557,9 +559,12 @@ public static function setHasBeenPaused($deliveryExecution, $paused)
         }
         /** @var DeliveryMonitoringService $deliveryMonitoringService */
         $deliveryMonitoringService = ServiceManager::getServiceManager()->get(DeliveryMonitoringService::SERVICE_ID);
-        $data = $deliveryMonitoringService->getData($deliveryExecution);
+
+        $data = new DeliveryMonitoringData($deliveryExecution, []);
```
hutnikau: `new DeliveryMonitoringData` statement used 17 times. I would like to not use `new` statement to instantiate this object, but use factory method as it was previously.

model/monitorCache/implementation/MonitoringStorage.php
jbout: the saveKvData() should probably be part of an else block, because the create includes the storing of keyvalue data

alexeyvy: @hutnikau that's intended, even when there is an exception, it means that the rows already exist and we just couldn't reveal them with update, as it returned 0 actually changed rows

hutnikau: If `create` call fails then `saveKvData` will fail too, so it does not make sense to catch those exceptions as it will definitely fail on next line of code. Maybe it makes sense to wrap `saveKvData` in `try` as well?

helpers/DeliveryHelper.php
alexeyvy: done

model/monitorCache/implementation/MonitoringStorage.php
alexeyvy: @jbout  The save/saveKv logic has been fixed
@hutnikau  Now the UniqueConstraintViolationException is used, so any general error will prevent executing saveKvData (that doesn't work if you use PDO instead of DBAL, because PDO has only the broad exception, but it's being fixed as per TAO-7005).

PR #644, opened 2018-10-30
Fix/tao 6758/proctor timer

tikhanovichA:


views/js/controller/Delivery/monitoring.js
```diff
@@ -923,6 +923,9 @@ define([
                                     remaining = 0;
```
zagovorichev: Check, please, that `rowTimer.consumedExtraTime` exists on the previous condition

zagovorichev:
Also, you can add  `qunit` tests for the case from this ticket

PR #647, opened 2018-11-06
TAO-7077 proctoring runner service fix

handleman:
move securePauseStateRequired param from getTestContext method to getTestData for proctor
-runner service

Related PR's:
https://github.com/oat-sa/extension-tao-test-runner-plugins/pull/82
https://github.com/oat-sa/extension-tao-act/pull/1080

### DIFF
--- a/model/runner/ProctoringRunnerService.php
+++ b/model/runner/ProctoringRunnerService.php
@@ -26,7 +26,6 @@ use oat\taoQtiTest\models\runner\QtiRunnerPausedException;
 use oat\taoQtiTest\models\runner\QtiRunnerService;
 use oat\taoQtiTest\models\runner\RunnerServiceContext;
 use qtism\runtime\tests\AssessmentTestSessionState;
-use oat\taoQtiTest\models\SectionPauseService;
 
 /**
  * Class ProctoringRunnerService
@@ -50,11 +49,11 @@ class ProctoringRunnerService extends QtiRunnerService
      */
     public function getTestData(RunnerServiceContext $context)
     {
-        $couldBePaused = $this->getServiceManager()->get(SectionPauseService::SERVICE_ID)->couldBePaused($context->getTestSession());
+        $testContext = parent::getTestContext($context);
         $testData = parent::getTestData($context);
 
-        if (isset($couldBePaused)) {
-            $testData['securePauseStateRequired'] = $couldBePaused;
+        if (isset($testContext['options']) && isset($testContext['options']['sectionPause'])) {
+            $testData['securePauseStateRequired'] = $testContext['options']['sectionPause'];
         } else {
             if ($context->getTestExecutionUri()) {
                 $deliveryExecution = ServiceProxy::singleton()->getDeliveryExecution($context->getTestExecutionUri());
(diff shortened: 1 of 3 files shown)


model/runner/ProctoringRunnerService.php
```diff
@@ -39,30 +39,31 @@ class ProctoringRunnerService extends QtiRunnerService
...
      * @return array
      * @throws \common_Exception
      * @throws \common_exception_NotFound
      * @throws \core_kernel_persistence_Exception
      */
-    public function getTestContext(RunnerServiceContext $context)
+    public function getTestData(RunnerServiceContext $context)
     {
-        $response = parent::getTestContext($context);
+        $testContext = parent::getTestContext($context);
```
krampstudio: Instead of retrieving the whole testContext, please call directly `$this->getServiceManager()->get(SectionPauseService::SERVICE_ID)->couldBePaused($context->getTestSession())` (or something like that) in order to get the value of `sectionPause`. 
And check if the test state is already in the interacting state  to support the proctoring implementation https://github.com/oat-sa/extension-tao-proctoring/blob/7cd8c4224a54ca00fb00ccef8bc820b14b7bd94c/model/execution/ProctoredSectionPauseService.php#L52

krampstudio:
@alexeyvy since you do the backend review, do you know how difficult it would be to test the method ?

alexeyvy:
> @alexeyvy since you do the backend review, do you know how difficult it would be to test the method ?

an hour for unit test (mocking getContext and getData and just testing our if-else conditions) and roughly several hours for testing the scenario (integration)

alexeyvy:
>I think we are at least in a feature change in semver since the data structure has changed.

I'd say we are in a major change, because this breaks backward compatibility

model/runner/ProctoringRunnerService.php
handleman: done in latest commit

handleman:
@alexeyvy @krampstudio  fix is pushed. I have changed feature version because API interface didn't change only additional functionality that is actually not more than new feature.

model/runner/ProctoringRunnerService.php
```diff
@@ -39,30 +40,31 @@ class ProctoringRunnerService extends QtiRunnerService
...
-    public function getTestContext(RunnerServiceContext $context)
+    public function getTestData(RunnerServiceContext $context)
     {
-        $response = parent::getTestContext($context);
+        $couldBePaused = $this->getServiceManager()->get(SectionPauseService::SERVICE_ID)->couldBePaused($context->getTestSession());
+        $testData = parent::getTestData($context);
 
-        if (isset($response['options']) && isset($response['options']['sectionPause'])) {
-            $response['securePauseStateRequired'] = $response['options']['sectionPause'];
+        if (isset($couldBePaused)) {
```
krampstudio: I'm having a doubt. 
May be this `if` was only there for backward compatibility and then I don't see how we will fall into the `else` since `canBePaused` will always have a value. 
@handleman could you please verify ; 
 - in tao community when are we in one case and when or we in the other
 - in tao with proctoring (act for example) when are we in one case and when or we in the other

I'm wondering if this condition was only there for backward compatibllity

handleman: After discussion with @alexeyvy I decided to revert changes to previous state where 
`if (isset($testContext['options']) && isset($testContext['options']['sectionPause']))`

presence of this parameter is checked instead of its value. and now it makes sense to check with `isset`

handleman: now `canBePaused` is checked this way: `if ($couldBePaused !== null) {`

handleman:
@krampstudio  all TAO-7077 PRs are not relevant since problem was fixed on PHP side. 
So maybe we can close PR without merging

PR #649, opened 2018-11-07
Feature/tao 7158/new task queue used

gyszucs:


zagovorichev:
@ionutpad FE part has been rewritten, it works and the PR can be reviewed

PR #662, opened 2019-02-01
Remove ambiguous default/continue in switch statement

cdorman:


peetya:
Thank you for submitting. Good catch! :)

PR #670, opened 2019-02-27
Feature/tao 7389/extend monitoring data with execution context

SashaTAO:
Store delivery execution context data in delivery monitoring data.

Requires: 
https://github.com/oat-sa/extension-tao-delivery/pull/385
https://github.com/oat-sa/extension-tao-lti/pull/174

### DIFF
--- a/model/monitorCache/implementation/DeliveryMonitoringData.php
+++ b/model/monitorCache/implementation/DeliveryMonitoringData.php
@@ -21,7 +21,8 @@
 
 namespace oat\taoProctoring\model\monitorCache\implementation;
 
-use oat\oatbox\service\ServiceManager;
+use oat\taoDelivery\model\execution\DeliveryExecutionContext;
+use oat\taoDelivery\model\execution\DeliveryExecutionContextInterface;
 use oat\taoDelivery\model\execution\DeliveryExecutionInterface;
 use oat\taoProctoring\model\execution\DeliveryExecutionManagerService;
 use oat\taoProctoring\model\implementation\TestSessionService;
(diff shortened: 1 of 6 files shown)


model/monitorCache/implementation/DeliveryMonitoringData.php
```diff
@@ -119,6 +120,34 @@ public function setDeliveryExecution(DeliveryExecution $deliveryExecution)
         $this->deliveryExecution = $deliveryExecution;
     }
 
+    /**
+     * @return DeliveryExecutionInterface
```
peetya: The PHPDoc of the object variable should be also updated to `DeliveryExecutionInterface`. What do you think?

model/monitorCache/implementation/DeliveryMonitoringData.php
```diff
@@ -119,6 +120,34 @@ public function setDeliveryExecution(DeliveryExecution $deliveryExecution)
...
+     */
+    public function getDeliveryExecution()
+    {
+        return $this->deliveryExecution;
+    }
+
+    /**
+     * {@inheritdoc
```
peetya: Missing `}` :)

model/monitorCache/implementation/DeliveryMonitoringData.php
```diff
@@ -119,6 +120,34 @@ public function setDeliveryExecution(DeliveryExecution $deliveryExecution)
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getDeliveryExecutionContext()
+    {
+        if (isset($this->data[self::PARAM_EXECUTION_CONTEXT])) {
+            return DeliveryExecutionContext::createFromArray(json_decode($this->data[self::PARAM_EXECUTION_CONTEXT], true));
+        }
+
+        return false;
```
peetya: What do you think about throwing an exception?

test/integration/monitorCache/mock/DeliveryMonitoringData.php
```diff
@@ -117,6 +120,34 @@ public function setDeliveryExecution(DeliveryExecution $deliveryExecution)
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getDeliveryExecutionContext()
+    {
+        if (isset($this->data[self::PARAM_EXECUTION_CONTEXT])) {
+            return DeliveryExecutionContext::createFromArray($this->data[self::PARAM_EXECUTION_CONTEXT]);
+        }
+
+        return false;
```
peetya: The same could be applied here.

SashaTAO: It's just a fake object for test.

model/monitorCache/implementation/DeliveryMonitoringData.php
SashaTAO: Fixed

model/monitorCache/implementation/DeliveryMonitoringData.php
SashaTAO: As we discussed it's not a required parameter and it will be better to return `null`

PR #678, opened 2019-03-20
 Contents on the clipboard are not cleared prior to launching an asse…

zagovorichev:
…ssment.

### DIFF
--- a/views/js/controller/DeliveryServer/awaiting.js
+++ b/views/js/controller/DeliveryServer/awaiting.js
@@ -54,30 +54,29 @@ define([
         var textToClipboard = ' ';
         var selected;
         var success;
-        var el;
+        var textAreaToSelContent;
         if (window.clipboardData) { // Internet Explorer
             window.clipboardData.setData ("Text", textToClipboard);
         } else {
-            el = document.createElement('textarea');  // Create a <textarea> element
-            loadingBar.stop();
-            el.setAttribute("id", "clipboardCleanerPlugin");
-            el.value = textToClipboard;                                 // Set its value to the string that you want copied
-            el.setAttribute('readonly', '');                // Make it readonly to be tamper-proof
-            el.style.position = 'absolute';
-            el.style.left = '-9999px';                      // Move outside the screen to make it invisible
-            document.body.appendChild(el);                  // Append the <textarea> element to the HTML document
+            textAreaToSelContent = document.createElement('textarea');  // Create a <textarea> element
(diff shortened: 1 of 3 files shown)


views/js/controller/DeliveryServer/awaiting.js
```diff
@@ -47,6 +47,44 @@ define([
...
+    var cleanClipboard = function cleanClipboard () {
+        var textToClipboard = ' ';
+        var selected;
+        var success;
+        var el;
+        if (window.clipboardData) { // Internet Explorer
+            window.clipboardData.setData ("Text", textToClipboard);
+        } else {
+            el = document.createElement('textarea');  // Create a <textarea> element
+            loadingBar.stop();
```
peetya: Why is it required to stop the loadingBar?

views/js/controller/DeliveryServer/awaiting.js
```diff
@@ -47,6 +47,44 @@ define([
      */
     var cssScope = '.awaiting-authorization';
 
+    /**
+     * Nothing needs to be taken to system clipboard on proceed action
+     */
+    var cleanClipboard = function cleanClipboard () {
```
peetya: This function sounds like a separate feature for me, and not something specific for the `DeliveryServer/awaiting.js`. It would make sense to create a global helper for that. What do you think?

views/js/controller/DeliveryServer/awaiting.js
```diff
@@ -47,6 +47,44 @@ define([
...
+        var textToClipboard = ' ';
+        var selected;
+        var success;
+        var el;
```
peetya: Could you please change the variable name to something more specific?

peetya:
Could you please cover the new function with unit tests?

views/js/controller/DeliveryServer/awaiting.js
zagovorichev: Not sure that we need this, maybe different library to control clipboards, but this won't be such a fast solution. (I guess that it will be a task for 8 story points with architect review, and for the core team).

zagovorichev:
@peetya I can't cover this method, because to test it I need real users action on the browser, which I can't do on the fake browsers. It could brake qunit tests

views/js/controller/DeliveryServer/awaiting.js
zagovorichev: yep, forgot to delete, I've moved this content from the test runner, good catch =)

zagovorichev:
@peetya Updated

PR #686, opened 2019-05-02
Fix/tao 8209 glitch in datatable

anshul119:
### Ticket: 

https://oat-sa.atlassian.net/browse/TAO-8209

### Summary

When opening the Proctor screen as an External user, the proctor screen loads lots of data to the Datagrid and then sorts the data.

### How to reproduce the issue?

Log in to the testing system as an external user on a day that no testing has occurred, and select a child organization (TU, ATEF, ATU). click the Proctor button. The page will display data and then will reset the table.

fnL91mxpUb

### How did I fix the issue?

Instead of selecting the default date range once the date picker is initialised I pass the default date range in the data table init config which removes the glitch.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -47,7 +47,7 @@ return array(
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '12.9.1',
+    'version' => '12.9.2',
     'author' => 'Open Assessment Technologies SA',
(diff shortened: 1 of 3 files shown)


krampstudio:
Closing since it has been merged to `develop` during https://github.com/oat-sa/extension-tao-proctoring/commit/b25a47d8e70bf7980240392b83b89b5dab0322c4 and this PR targets master.

PR #712, opened 2019-07-02
added DeliveryLogEvent, to know when records added to the delivery_log

zagovorichev:
Pull request summary

Related to: https://oat-sa.atlassian.net/browse/TAO-8585

New events added to have a possibility to listen when delivery_log added:

added *DeliveryLogEvent*

How to test
You can test it with the taoSyncClient extension (https://github.com/oat-sa/extension-tao-sync-client/tree/develop)
taoSyncClient creates sync_client_queue table and add new rows on delivery_log insert action


model/deliveryLog/event/DeliveryLogEvent.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @author Oleksandr Zagovorychev <zagovorichev@gmail.com>
+ */
+
+namespace oat\taoProctoring\model\deliveryLog\event;
+
+
+use oat\oatbox\event\Event;
+
+class DeliveryLogEvent implements Event
```
SashaTAO: What do you think, would it make sense to rename to DeliveryLogCreatedEvent?

zagovorichev: I guess no because the method that triggers it calls `log`. So if we get it from the class interface we don't know what exactly happens there.

Yury-Razhkou:
Tested.Works as described.

PR #714, opened 2019-07-03
Feature/tao 8556/remove unecessary columns

SergiiTao:
Ticket: https://oat-sa.atlassian.net/browse/TAO-8556

Added possibility to hide columns in the monitoring dashboard


model/GuiSettingsService.php
```diff
@@ -49,8 +49,31 @@ class GuiSettingsService extends ConfigurableService
...
+            self::OPTION_SHOW_COLUMN_FIRST_NAME     => $this->hasOption(self::OPTION_SHOW_COLUMN_FIRST_NAME) ? $this->getOption(self::OPTION_SHOW_COLUMN_FIRST_NAME) : true,
+            self::OPTION_SHOW_COLUMN_LAST_NAME      => $this->hasOption(self::OPTION_SHOW_COLUMN_LAST_NAME) ? $this->getOption(self::OPTION_SHOW_COLUMN_LAST_NAME) : true,
+            self::OPTION_SHOW_COLUMN_AUTHORIZE      => $this->hasOption(self::OPTION_SHOW_COLUMN_AUTHORIZE) ? $this->getOption(self::OPTION_SHOW_COLUMN_AUTHORIZE) : true,
+            self::OPTION_SHOW_COLUMN_REMAINING_TIME => $this->hasOption(self::OPTION_SHOW_COLUMN_REMAINING_TIME) ? $this->getOption(self::OPTION_SHOW_COLUMN_REMAINING_TIME) : true,
+            self::OPTION_SHOW_COLUMN_EXTENDED_TIME  => $this->hasOption(self::OPTION_SHOW_COLUMN_EXTENDED_TIME) ? $this->getOption(self::OPTION_SHOW_COLUMN_EXTENDED_TIME) : true,
+            self::OPTION_SHOW_COLUMN_CONNECTIVITY   => $this->hasOption(self::OPTION_SHOW_COLUMN_CONNECTIVITY) ? $this->getOption(self::OPTION_SHOW_COLUMN_CONNECTIVITY) : true,
```
bziondik: I think 'true' is not the right default value,  because of
https://github.com/oat-sa/extension-tao-proctoring/blob/e4ee3e2b9c2f327d14ad7c50719a338900a375bf/views/js/controller/Delivery/monitoring.js#L619

model/GuiSettingsService.php
```diff
@@ -49,8 +49,31 @@ class GuiSettingsService extends ConfigurableService
...
+
+    /**
+     * @return array
+     */
+    public function asArray()
     {
-        parent::__construct($options);
+        return [
+            self::PROCTORING_REFRESH_BUTTON         => $this->hasOption(self::PROCTORING_REFRESH_BUTTON) ? $this->getOption(self::PROCTORING_REFRESH_BUTTON) : true,
+            self::PROCTORING_AUTO_REFRESH           => $this->hasOption(self::PROCTORING_AUTO_REFRESH) ? $this->getOption(self::PROCTORING_AUTO_REFRESH) : true,
```
bziondik: here default value '0'
https://github.com/oat-sa/extension-tao-proctoring/blob/e4ee3e2b9c2f327d14ad7c50719a338900a375bf/scripts/install/RegisterGuiSettingsService.php#L45

model/GuiSettingsService.php
uncleempty: This is right default value.
Both BE and FE are using boolean value in that case.

views/js/controller/Delivery/monitoring.js
```diff
@@ -597,12 +597,17 @@ define([
                     return `${moment().format(dateFormat)} to ${moment().add('1', 'd').format(dateFormat)}`;
                 }
 
+                function extractOption(object, option, defaultValue) {
+                    return Object.is(object[option], undefined) ? defaultValue : object[option];
```
uncleempty: One small question:
is `Object.is` available in ES5? Because, as I remember, it was still a draft, and introduced only in ES6. And, since, some of the customers may possibly use older sprints (96), which not fully ES6-compliant, - would everything be ok while using it?
@bziondik , could you help with that question?

uncleempty: Wouldn't it be better to use smth like `_.isUndefined(object[option]) ? defaultValue : object[option]`?

model/GuiSettingsService.php
bziondik: it is not about boolean, it is about logic
On frontend, if we don't have onlineStatus then it will be false (allowedConnectivity = data.onlineStatus || false)
On backend,  if we don't have onlineStatus then it will be true (self::OPTION_SHOW_COLUMN_CONNECTIVITY   => $this->hasOption(self::OPTION_SHOW_COLUMN_CONNECTIVITY) ? $this->getOption(self::OPTION_SHOW_COLUMN_CONNECTIVITY) : true)

views/js/controller/Delivery/monitoring.js
bziondik: @uncleempty  taoProctoring is migrated on ES2015(ES6) https://oat-sa.atlassian.net/wiki/spaces/NEX/pages/78971530/Migration+to+ES2015
We can use:
`Object` and `Array` new built-in (`Objet.assign`, `Array.from`, `Array.of`, etc.)

model/GuiSettingsService.php
SergiiTao: fixed

views/js/controller/Delivery/monitoring.js
uncleempty: Yep, this question is mainly about next situation: proctoring is migrated, but I'm not sure at all that we would move DEPP's UAT/prod to more recent sprints than 96 (and migration to ES6 was later), at least it wasn't discussed yet. That's why the question was raised.

SergiiTao: @jsconan what would you suggest?

jsconan: For a backport, indeed it could be an issue, as we won't be able to transpile it.
However, for new development this is totally legit since the extension is turned into ES6 compatible (`babel` flag in the `bundle.js`)

PR #716, opened 2019-07-15
Fix/tao 8746

tikhanovichA:
https://oat-sa.atlassian.net/browse/TAO-8746


helpers/DeliveryHelper.php
```diff
@@ -377,7 +378,11 @@ private static function adjustDeliveryExecutions($deliveryExecutions) {
             $testTaker['test_taker_first_name'] = (isset($cachedData[DeliveryMonitoringService::TEST_TAKER_FIRST_NAME]))?_dh($cachedData[DeliveryMonitoringService::TEST_TAKER_FIRST_NAME]):'';
 
             foreach(self::_getUserExtraFields() as $field){
-                $extraFields[$field['id']] = isset($cachedData[$field['id']]) ? _dh($cachedData[$field['id']]) : '';
+                $value = isset($cachedData[$field['id']]) ? _dh($cachedData[$field['id']]) : '';
+                if (\common_Utils::isUri($value)) {
```
bartlomiejmarszal: We are changing logic here, we should add test for that

PR #721, opened 2019-07-30
Register DeliveryExecutionFinished event in webhooks service

siarhei-baradzin:
https://oat-sa.atlassian.net/browse/TAO-8497

Depends on: https://github.com/oat-sa/tao-core/pull/2235
Register DeliveryExecutionFinished event in webhooks service


manifest.php
```diff
@@ -47,10 +48,10 @@
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '16.4.0',
+    'version' => '16.4.1',
     'author' => 'Open Assessment Technologies SA',
     'requires' => array(
-        'tao'            => '>=36.0.0',
+        'tao'            => '>=38.3.0',
```
tikhanovichA: if you change a major version of a dependency, then you should up the same version of this extension

siarhei-baradzin: thx, fixed

PR #728, opened 2019-08-26
Fix/tao 8599/fix progress translations

tikhanovichA:
Depends on https://github.com/oat-sa/extension-tao-testqti/pull/1581


helpers/DeliveryHelper.php
```diff
@@ -364,9 +364,19 @@ private static function adjustDeliveryExecutions($deliveryExecutions) {
         $executions = [];
         foreach ($deliveryExecutions as $cachedData) {
 
+            $progressPatern = "/(?<=\- )(.*?)(?= \d)/";
```
alexeyvy: Having a test case covering this logic wouldn't be out of place

tikhanovichA: It's should be done not in this PR. Also, it's the private static method from helper, and it brings a lot of complexity if we want to create a unit test exactly for that part

tikhanovichA: Do you agree?

alexeyvy: ok

PR #730, opened 2019-09-04
Progress field incorrect

bartlomiejmarszal:
Class `\oat\taoProctoring\helpers\DeliveryHelper` is based on static methods. To repair this bug I decided to move functionality that was used to generate field where bug was spotted, into new service. I created new class `DeliveryHelperService` that contains `public function adjustDeliveryExecutions` that is tested and available thru `ServiceLocator`

https://oat-sa.atlassian.net/browse/TAO-8997


helpers/DeliveryHelper.php
```diff
@@ -39,6 +38,7 @@
 use tao_helpers_Date as DateHelper;
 
 /**
+ * @deprecated
```
tikhanovichA: Could you please add some info about what we should use now?

siarhei-baradzin:
Are you going to remove old implementation from DeliveryHelper?

bartlomiejmarszal:
> Are you going to remove old implementation from DeliveryHelper?

Yes. I'm working on that now

manifest.php
```diff
@@ -48,7 +48,7 @@
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '17.2.2',
+    'version' => '17.2.3',
```
peetya: Deprecating classes and removing public methods doesn't seem like a PATCH version change.

bartlomiejmarszal: I didn't change any behaviour, business logic and class did not lost any dependencies. It was `just` refactoring so it's just a patch that is contained in [132b0141391e46b30869e9af051678f5f82d6202](https://github.com/oat-sa/extension-tao-proctoring/pull/730/commits/132b0141391e46b30869e9af051678f5f82d6202)

helpers/DeliveryHelperService.php
```diff
@@ -0,0 +1,287 @@
...
+use oat\taoProctoring\model\execution\DeliveryExecution;
+use oat\taoProctoring\model\monitorCache\DeliveryMonitoringService;
+use oat\taoProctoring\model\TestSessionConnectivityStatusService;
+use oat\taoQtiTest\models\SessionStateService;
+
+/**
+ * Class DeliveryHelperService
+ * @author Bartlomiej Marszal
+ */
+class DeliveryHelperService extends ConfigurableService
```
Skyld: So, based on what we discussed in person, here are a few suggestions how to apply the boy scout rule (https://medium.com/@biratkirat/step-8-the-boy-scout-rule-robert-c-martin-uncle-bob-9ac839778385) on this task :)

- Use Cover and modify approach before refactoring (https://docs.google.com/presentation/d/1lnmxPT50g8C8yeAnE_lnQWjJDshO1o6jWWk6Kc6T1zM/edit#slide=id.g596cf22888_0_96)  
- Try to get rid of the ServiceManager, inject the actual dependencies via constructor
- Try to wrap static helper classes with other objects and inject them as in constructor
- Reduce Npath and cyclomatic complexity of the methods (https://modess.io/npath-complexity-cyclomatic-complexity-explained/)
-  Try to enforce the object calisthenics rules (one level indentation per method, avoid ELSE clauses, etc... https://williamdurand.fr/2013/06/03/object-calisthenics/)
- Try to apply single responsibility principle, this class does way too many things
- Get rid of the inconsistent return points of your methods (true|false|null, etc)
- Try to introduce value objects instead of working with plain arrays

I think this is enough for now :)

bartlomiejmarszal: I apply most of your points into my changes. Some of them were to time consuming and I couldn't perform them because it's only bug fix

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /**
+     * @param $cachedData
+     * @param $userExtraFields
+     * @return array
+     * @throws \common_exception_Error
+     * @throws \common_ext_ExtensionException
+     */
+    private function getSingleExecution($cachedData, $userExtraFields)
```
julien-sebire: This function is still a big thing but nice with only one complexity level.
If you want to go further for something way clearer, I'd split it in parts:
- line 70 to 75: create the state
- move line 78 before the `foreach` in line 89 where it belongs to
- lines 77 to 87: create the test taker
- lines 88 to 92: create the extra fields
- lines 93 to 129: create the execution

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+        $elapsedApprox = 0;
+        $executionState = $cachedData[DeliveryMonitoringService::STATUS];
+
+        if (
+            $executionState === DeliveryExecution::STATE_ACTIVE
+            && isset($cachedData[DeliveryMonitoringService::LAST_TEST_TAKER_ACTIVITY])
+        ) {
+            $lastActivity = $cachedData[DeliveryMonitoringService::LAST_TEST_TAKER_ACTIVITY];
+            $elapsedApprox = $now - $lastActivity;
+            $duration = isset($cachedData[DeliveryMonitoringService::ITEM_DURATION])
```
julien-sebire: Can we use `??` operator with (float) casting outside?

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+        return round((float)$remaining - $elapsedApprox);
+    }
+
+    /**
+     * @param array $cachedData
+     * @return int
+     */
+    private function getRemainingTime(array $cachedData)
+    {
+        $remaining = isset($cachedData[DeliveryMonitoringService::REMAINING_TIME])
```
julien-sebire: Can we use `??` operator with (int) casting outside?

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+
+    /**
+     * Get array of user specific extra fields to be displayed in the monitoring data table
+     *
+     * @return array
+     * @throws \common_ext_ExtensionException
+     */
+    private function getUserExtraFields()
+    {
+        $extraFields = [];
```
julien-sebire: This line can move above the `foreach` in line 216.

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+            $extraFields[] = $this->mergeExtraFieldsSettings($uri, $name, $userExtraFieldsSettings);
+        }
+
+        return $extraFields;
+    }
+
+    private function mergeExtraFieldsSettings($uri, $name, $userExtraFieldsSettings)
+    {
+        $property = $this->getProperty($uri);
+        $settings = array_key_exists($name, $userExtraFieldsSettings)
```
julien-sebire:
```
array_key_exists($name, $userExtraFieldsSettings)
```
is the same as
```
isset($userExtraFieldsSettings[$name])
```
Which can then lead to:
```
$settings = $userExtraFieldsSettings[$name] ?? [];
```

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+
+    /**
+     * @param array $cachedData
+     * @return mixed|string
+     */
+    private function getProgressString(array $cachedData)
+    {
+        $progressStr = $cachedData[DeliveryMonitoringService::CURRENT_ASSESSMENT_ITEM];
+
+        if (($progress = json_decode($progressStr, true)) === null) {
```
julien-sebire: Please split this in two lines, for the sake of reading. YOu already have the json decoded into a variable, no need to put it on the same line:
```
$progress = json_decode($progressStr, true);
if ($progress === null) {
```

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+        }
+        $format = $this->getSessionStateService()->hasOption(SessionStateService::OPTION_STATE_FORMAT)
+            ? $this->getSessionStateService()->getOption(SessionStateService::OPTION_STATE_FORMAT)
+            : __('%s - item %p/%c');
+        $map = array(
+            '%s' => $progress['title'] ?? '',
+            '%p' => $progress['itemPosition'] ?? '',
+            '%c' => $progress['itemCount'] ?? ''
+        );
+        $progressStr = strtr($format, $map);
```
julien-sebire: You can directly return the value instead of putting it into a variable.

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+        return $progressStr;
+    }
+
+    /**
+     * @param array $cachedData
+     * @return bool|null
+     */
+    private function isOnline(array $cachedData)
+    {
+        $online = null;
```
julien-sebire: Would do:
```
        if ($this->getTestSessionConnectivityStatusService()->hasOnlineMode()) {
            return $this->getTestSessionConnectivityStatusService()->isOnline($cachedData[DeliveryMonitoringService::DELIVERY_EXECUTION_ID]);
        }
        return null;
```

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /**
+     * @param array $cachedData
+     * @param bool|null $online
+     * @return float|null
+     */
+    private function getLastActivity(array $cachedData, ?bool $online)
+    {
+        if ($online && isset($cachedData[DeliveryMonitoringService::LAST_TEST_TAKER_ACTIVITY])) {
```
julien-sebire: Remove the `else`:
```
if ($online && isset($cachedData[DeliveryMonitoringService::LAST_TEST_TAKER_ACTIVITY])) {
    return $cachedData[DeliveryMonitoringService::LAST_TEST_TAKER_ACTIVITY];
}
return null;
```

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /**
+     * @return common_ext_ExtensionsManager
+     */
+    private function getExtensionManagerService()
+    {
+        return $this->getServiceLocator()->get(common_ext_ExtensionsManager::SERVICE_ID);
+    }
+}
```
julien-sebire: Missing newline at end of file.

model/execution/DeliveryExecutionList.php
bartlomiejmarszal: > lines 88 to 92: create the extra fields

Should I extract foreach to separate method?

helpers/DeliveryHelperService.php
Skyld: OK I'll approve, there is no need to push this further in scope of this bugfix. Nice job 👍

model/execution/DeliveryExecutionList.php
```diff
@@ -0,0 +1,336 @@
...
+
+    /**
+     * @param $input
+     * @return string
+     * @throws \common_exception_Error
+     * @throws \common_ext_ExtensionException
+     */
+    private function sanitizeUserInput($input)
+    {
+        return htmlentities($input, ENT_COMPAT, $this->getApplicationService()->getDefaultEncoding());
```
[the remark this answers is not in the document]
bartlomiejmarszal: No. This will be long story until I will get this settle so I would like to skip this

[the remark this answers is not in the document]
julien-sebire: Ok, this will stay as is for long I guess then... :)

PR #735, opened 2019-09-17
Fix broken DeliveryExecutionListTest

bartlomiejmarszal:


alexeyvy:
version bump?

test/unit/model/execution/DeliveryExecutionListTest.php
```diff
@@ -210,8 +210,8 @@ public function testAdjustDeliveryExecutionsProgressStringWithNoOption()
 
         $this->sessionStateServiceMock->method('hasOption')->willReturn(false);
         $this->proctoringExtensionMock->method('getConfig')->willReturn(null);
-        /** @var Model|PHPUnit_Framework_MockObject_MockObject $modelMock */
-        $modelMock = $this->createMock(Model::class);
+        /** @var Ontology|PHPUnit_Framework_MockObject_MockObject $modelMock */
+        $modelMock = $this->createMock(Ontology::class);
```
alexeyvy: it's already created from Ontology::class in the method setUp:

`$this->modelMock = $this->createMock(Ontology::class);`

so you can just delete this line and proceed with `$this->modelMock ` instead of `$modelMock` in the code below

PR #738, opened 2019-10-09
feat: add new language (lt-LT)

AndreyNegrov:
add new language (lt-LT)


tikhanovichA:
Is it ok, that I see lines without translates?

locales/lt-LT/messages.po
```diff
@@ -0,0 +1,1007 @@
...
+msgid "%s minutes more"
+msgstr "%s minučių daugiau"
+
+msgid "%s\'s test had already been authorized."
+msgstr ""
+
+msgid "%s\'s test had already been paused."
+msgstr ""
+
+msgid "%s\'s test had already been terminated."
```
tikhanovichA: As I remember it will not work if you use backslashes

AndreyNegrov: yes, you right. it fixed

PR #753, opened 2019-12-24
add ACTIVE delivery execution state to state check

yuryfilippovich:
Related with: https://oat-sa.atlassian.net/browse/TAO-9572

Steps to reproduce:
1 Create proctored and not secure delivery 
2 Log in as a test taker and run delivery
3 Activate test taker session as proctor
4 Start test as test taker and in any step refresh page
The same steps  for lti run
Expected result: test-taker able to continue test
Actual result: test-taker session become paused and request a proctor activation again

### DIFF
--- a/model/authorization/TestTakerAuthorizationService.php
+++ b/model/authorization/TestTakerAuthorizationService.php
@@ -121,7 +121,7 @@ class TestTakerAuthorizationService extends ConfigurableService implements TestT
      */
     public function isActiveUnSecureDelivery($deliveryId, $state)
     {
-        return $this->isNotSecure($deliveryId) && $state === DeliveryExecutionInterface::STATE_ACTIVE;
+        return $state === DeliveryExecutionInterface::STATE_ACTIVE && !$this->isSecure($deliveryId);
     }
 
     /**
@@ -129,7 +129,7 @@ class TestTakerAuthorizationService extends ConfigurableService implements TestT
      * @return bool
      * @throws common_Exception
      */
-    private function isNotSecure($deliveryId)
+    private function isSecure($deliveryId)
     {
         $delivery = $this->getResource($deliveryId);
         $activeFeatures = explode(
@@ -138,7 +138,7 @@ class TestTakerAuthorizationService extends ConfigurableService implements TestT
                 $this->getProperty(DeliveryContainerService::TEST_RUNNER_FEATURES_PROPERTY)
             )
         );
-        return !in_array('security', $activeFeatures);
+        return in_array('security', $activeFeatures);
     }
 
     /**
--- a/model/authorization/TestTakerAuthorizationService.php
+++ b/model/authorization/TestTakerAuthorizationService.php
@@ -19,8 +19,10 @@
  */
 namespace oat\taoProctoring\model\authorization;
 
-use oat\oatbox\event\EventManager;
+
+use common_Exception;
 use oat\oatbox\service\ConfigurableService;
+use oat\taoDeliveryRdf\model\DeliveryContainerService;
 use oat\taoDelivery\model\execution\DeliveryExecution;
 use oat\taoDelivery\model\execution\DeliveryExecutionInterface;
 use oat\taoProctoring\model\DelegatedServiceHandler;
@@ -70,6 +72,7 @@ class TestTakerAuthorizationService extends ConfigurableService implements TestT
     public function verifyResumeAuthorization(DeliveryExecutionInterface $deliveryExecution, User $user)
     {
         $state = $deliveryExecution->getState()->getUri();
+        $deliveryUri= $deliveryExecution->getDelivery()->getUri();
         if (in_array($state, [
             ProctoredDeliveryExecution::STATE_FINISHED,
             ProctoredDeliveryExecution::STATE_CANCELED,
@@ -80,12 +83,11 @@ class TestTakerAuthorizationService extends ConfigurableService implements TestT
                 'Terminated/Finished delivery execution "'.$deliveryExecution->getIdentifier().'" cannot be resumed'
             );
         }
+
         if (
-            $this->isProctored($deliveryExecution->getDelivery()->getUri(), $user)
-            && !in_array(
-                $state,
-                [ProctoredDeliveryExecution::STATE_AUTHORIZED, ProctoredDeliveryExecution::STATE_ACTIVE]
-            )
+            $this->isProctored($deliveryUri, $user)
+            && $state !== ProctoredDeliveryExecution::STATE_AUTHORIZED
+            && !$this->isActiveUnSecureDelivery($deliveryUri, $state)
         ) {
             $this->throwUnAuthorizedException($deliveryExecution);
         }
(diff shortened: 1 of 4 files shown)


model/authorization/TestTakerAuthorizationService.php
```diff
@@ -80,7 +80,13 @@ public function verifyResumeAuthorization(DeliveryExecutionInterface $deliveryEx
                 'Terminated/Finished delivery execution "'.$deliveryExecution->getIdentifier().'" cannot be resumed'
             );
         }
-        if ($this->isProctored($deliveryExecution->getDelivery()->getUri(), $user) && $state !== ProctoredDeliveryExecution::STATE_AUTHORIZED) {
+        if (
```
SashaTAO: Could you please write unit tests for this method?

model/authorization/TestTakerAuthorizationService.php
```diff
@@ -104,6 +113,34 @@ public function isProctored($deliveryId, User $user)
...
+    {
+        return $this->isNotSecure($deliveryId) && $state === DeliveryExecutionInterface::STATE_ACTIVE;
+    }
+
+    /**
+     * @param string $deliveryId
+     * @return bool
+     * @throws common_Exception
+     */
+    private function isNotSecure($deliveryId)
```
SashaTAO: Could you please rename it to `isSecure` or `isSecureDelivery` and use `!isSecure()` in a place where you call this method.

model/authorization/TestTakerAuthorizationService.php
```diff
@@ -104,6 +113,34 @@ public function isProctored($deliveryId, User $user)
...
 
+    /**
+     * @param string $deliveryId
+     * @param string $state
+     * @return bool
+     * @throws common_Exception
+     */
+    public function isActiveUnSecureDelivery($deliveryId, $state)
+    {
+        return $this->isNotSecure($deliveryId) && $state === DeliveryExecutionInterface::STATE_ACTIVE;
```
SashaTAO: It would be better in terms of performance to check constant value at first to do not call DB if $state is not equal `DeliveryExecutionInterface::STATE_ACTIVE`

test/unit/model/authorization/TestTakerAuthorizationServiceTest.php
```diff
@@ -0,0 +1,300 @@
...
+            ->willReturn($property);
+
+        $delivery->expects($this->once())
+            ->method('getOnePropertyValue')
+            ->with($property)
+            ->willReturn($propertyValue);
+
+        $this->service->setServiceLocator($this->getServiceLocatorMock([Ontology::SERVICE_ID => $this->ontologyMock]));
+
+        if ($expected) {
```
SashaTAO: It's not really a good practice to have logic like `if() {} else {}` in your test. It would be better to check `$this->assertEquals($expected, $result);` or have two different test cases.

test/unit/model/authorization/TestTakerAuthorizationServiceTest.php
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    /**
+     * @return array
+     */
+    public function isActiveUnSecureDeliveryDataProvider()
+    {
+        return [
+            'activeAndUnSecure' => [
+                null, 'http://www.tao.lu/Ontologies/TAODelivery.rdf#DeliveryExecutionStatusActive', true
```
SashaTAO: If you use names in dataProvider it's easier to understand your test:
```
'activeAndUnSecure' => [
    'propertyValue' => null,
    'executionState' => 'http://www.tao.lu/Ontologies/TAODelivery.rdf#DeliveryExecutionStatusActive',
    'expectedResult' => true
],
```

test/unit/model/authorization/TestTakerAuthorizationServiceTest.php
```diff
@@ -0,0 +1,300 @@
...
+        $this->service->setServiceLocator(
+            $this->getServiceLocatorMock(
+                [
+                    Ontology::SERVICE_ID => $this->ontologyMock,
+                    DeliverySyncService::SERVICE_ID => $deliverySyncServiceMock
+                ]
+            )
+        );
+
+        if ($expected) {
```
SashaTAO: The same, it's better to check `$this->assertEquals($expected, $result);` instead of using `if`

test/unit/model/authorization/TestTakerAuthorizationServiceTest.php
```diff
@@ -0,0 +1,300 @@
...
+
+        $this->service->setServiceLocator(
+            $this->getServiceLocatorMock(
+                [
+                    Ontology::SERVICE_ID => $this->ontologyMock,
+                    DeliverySyncService::SERVICE_ID => $this->getMock(DeliverySyncService::class)
+                ]
+            )
+        );
+        if (!$isSuccess) {
```
SashaTAO: If you test than method throws exception it's better to have it as a separate test case where you will test all possible scenarios with exception.

model/authorization/TestTakerAuthorizationService.php
yuryfilippovich: fixed

test/unit/model/authorization/TestTakerAuthorizationServiceTest.php
yuryfilippovich: fixed

model/authorization/TestTakerAuthorizationService.php
yuryfilippovich: changed

test/unit/model/authorization/TestTakerAuthorizationServiceTest.php
yuryfilippovich: fixed

test/unit/model/authorization/TestTakerAuthorizationServiceTest.php
yuryfilippovich: fixed

test/unit/model/authorization/TestTakerAuthorizationServiceTest.php
yuryfilippovich: fixed

zagovorichev:
> These changes fix reported problem. But if "Secure mode" is on, after refresh delivery execution is not paused as well and doesn't require authorization by proctor. Could you please clarify with QA if it's expected behavior?

@Nataliame Could you confirm this behavior, please?

model/authorization/TestTakerAuthorizationService.php
```diff
@@ -104,6 +113,34 @@ public function isProctored($deliveryId, User $user)
...
+    private function isSecure($deliveryId)
+    {
+        $delivery = $this->getResource($deliveryId);
+        $activeFeatures = explode(
+            ',',
+            $delivery->getOnePropertyValue(
+                $this->getProperty(DeliveryContainerService::TEST_RUNNER_FEATURES_PROPERTY)
+            )
+        );
+        return in_array('security', $activeFeatures);
```
zagovorichev: Could you use const instead of string ?  `SecurityFeature::FEATURE_ID`

yuryfilippovich: class SecurityFeature marked us deprecated by @hutnikau. I discussed it with him and we decided use string instead constant

PR #756, opened 2020-01-15
WTP-1512 / TSTDEL-940 - Remove the ability to enter irr

alan92rttt:
WTP-1512 / TSTDEL-940 - Remove the ability to enter irr

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -48,7 +48,7 @@ return array(
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '18.2.0',
+    'version' => '18.2.1',
     'author' => 'Open Assessment Technologies SA',
     'requires' => array(
         'tao'            => '>=38.9.0',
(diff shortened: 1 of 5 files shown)


zagovorichev:
Hello @alan92rttt 
1. You have to make these changes in the extension-tao-act repo, as this is ACT feature.
2. Also, to implement new lti parameter you have to use custom lti parameters. (example https://github.com/oat-sa/extension-lti-proctoring/blob/b8e260fab4dacd993b255a4f181c8cd928d895f0/model/delivery/LtiTestTakerAuthorizationService.php#L43)

FYI:
you can test it with http://lti.tools/test/tc.php

zagovorichev:
3. please, assign me to your new pull request in the `extension-tao-act` (then I will know that there are changes and I can check it)
4. Do not forget to write PHPUnit tests, please

alan92rttt:
@zagovorichev , The fact that this change is not in the ACT extension was discussed prior to the creation of this PR. Here is a comment from the slack discussion. 


Oleksandr Khitko  9:21 AM
We checked it yesterday and there is no easy way to do this chenge  in ACT custom extension without massive code duplication or changes in taoProctoring.

But we can implement it in taoProctoring with default behavior as it is now and possibility to turn off irregularity button via LTI parameter

It will not affect any of existing instances and add more flexibility

zagovorichev:
@alan92rttt , If I understood this correctly you can change frontend part in the taoProctoring, but not the way you did.

Suggestion:
you could try to get lti parameters for the defined delivery execution, instead of collecting everything for all data (is a very heavy load on the server)

and for the backend part: everything has to be implemented in another extension - taoAct
(definition of new lti parameter and detection that we don't need to show irregularities button)

PR #769, opened 2020-02-12
Feature/actp 245/proctor remaining time

augustas:
Make remaining time on proctor screen more accurate after pause event.
 
Related to : https://oat-sa.atlassian.net/browse/ACTP-245
 
This PR adds an additional value stored in delivery monitoring cache, to track the time that was already stored in timer duration storage.

Steps to reproduce the initial bug and how to test is described in companion PR: https://github.com/oat-sa/extension-tao-act/pull/1295
 

#### Dependencies
 
Companion :
 - [ ] https://github.com/oat-sa/extension-tao-act/pull/1295

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -48,7 +48,7 @@ return array(
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '18.2.5',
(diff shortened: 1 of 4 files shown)


zagovorichev:
@augustas there is a warning, could you check it , please? :

PR #778, opened 2020-03-16
Fix/tao 371/restyle start resume buttons

AnastasiaArcadia:
**Related to:**
https://oat-sa.atlassian.net/browse/ACTP-371

**Description:**
Design of buttons on Delivery page, and on Proctoring page are different at the moment.

**How to check:**
Create test with proctoring required.
Login as test taker.
Delivery page and Proctoring page should have similar style.


jsconan:
Pay also attention naming the branch with the correct ticket: `TAO-371` is not the same as `ACTP-371`.

PR #782, opened 2020-03-19
Fix/actp 371/restyle start resume buttons

AnastasiaArcadia:
**Related to:**
https://oat-sa.atlassian.net/browse/ACTP-371

**Description:**
Design of buttons on Delivery page, and on Proctoring page are different at the moment.

**How to check:**
Create test with proctoring required.
Login as test taker.
Delivery page and Proctoring page should have similar style.


manifest.php
```diff
@@ -48,7 +48,7 @@
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '19.2.1',
+    'version' => '19.2.2',
     'author' => 'Open Assessment Technologies SA',
     'requires' => array(
         'tao'            => '>=38.9.0',
```
jsconan: You should update this line as well, since the fix is relying on a packaged installed by taoCore in `41.7.1`

PR #792, opened 2020-04-07
Feature/actp 394/timer adjustment by proctor

augustas:
Time adjustment endpoint for proctor.
 
Related to : https://oat-sa.atlassian.net/browse/ACTP-394
 
New endpoint in monitoring, to store forward the timer adjustment.
  
#### Dependencies
 
Requires :
 - [x] https://github.com/oat-sa/extension-tao-test/pull/321
 - [x] https://github.com/oat-sa/extension-tao-testqti/pull/1726


model/execution/DeliveryExecutionManagerService.php
```diff
@@ -238,4 +236,60 @@ public function setExtraTime($deliveryExecutions, $extraTime = 0, $extendedTime
...
+                    $success = $timerAdjustmentService->increase($deliveryExecution, $seconds);
+                } else {
+                    $success = $timerAdjustmentService->decrease($deliveryExecution, abs($seconds));
+                }
+            }
+
+            if ($success) {
+                $result['processed'][$deliveryExecution->getIdentifier()] = true;
+            } else {
+                $result['unprocessed'][$deliveryExecution->getIdentifier()] = false;
```
zagovorichev, line 264: why uprocessed === false is true? why unprocessed = true is not unprocessed?

model/implementation/DeliveryExecutionStateService.php
```diff
@@ -57,6 +57,9 @@ class DeliveryExecutionStateService extends AbstractStateService implements \oat
     const OPTION_CANCELLATION_DELAY = 'cancellation_delay';
     const OPTION_TIME_HANDLING = 'time_handling';
 
+    const TIME_HANDLING_EXTRA_TIME = 'extra_time';
```
zagovorichev, line 60: I'm just wondering : what if we rename this variable https://github.com/oat-sa/extension-tao-proctoring/blob/8b98ae91a1ce06954d9456ad34ad1cabaee3784a/model/monitorCache/DeliveryMonitoringService.php#L55, then this one should be renamed as well? then I prefer only 1 constant

model/execution/DeliveryExecutionManagerService.php
```diff
@@ -238,4 +236,60 @@ public function setExtraTime($deliveryExecutions, $extraTime = 0, $extendedTime
...
+        }
+
+        return $result;
+    }
+
+    /**
+     * @param DeliveryExecutionInterface $deliveryExecution
+     * @return bool
+     */
+    private function isCorrectDeliveryExecutionStateForAdjustment($deliveryExecution)
```
SashaTAO, line 275: Maybe we could rename it to something shorter like `canAdjustTimer` or `timerAdjustmentAllowed`. We may have some restrictions not related to state later.

model/implementation/DeliveryExecutionStateService.php
```diff
@@ -57,6 +57,9 @@ class DeliveryExecutionStateService extends AbstractStateService implements \oat
...
+    const TIME_HANDLING_TIMER_ADJUSTMENT = 'timer_adjustment';
```
SashaTAO, line 61: Why do we need these constants if they are not used?

model/execution/DeliveryExecutionManagerService.php
augustas: Current frontend implementation expects such values, same as in `DeliveryExecutionManagerService::setExtraTime()`. I agree it's not very clear, maybe original developer though `unprocessed` is a failure and should get a `false`.

model/implementation/DeliveryExecutionStateService.php
augustas: The constant you've linked is concerning a key to use when storing data in `delivery_monitoring` table. While the new constant in this file is for the values valid for option `OPTION_TIME_HANDLING`

model/execution/DeliveryExecutionManagerService.php
augustas: Done

model/implementation/DeliveryExecutionStateService.php
augustas: I've updated the updater to use the constant `TIME_HANDLING_EXTRA_TIME`, and client extensions which require the timer adjustment feature will use the second constant `TIME_HANDLING_TIMER_ADJUSTMENT`.

zagovorichev:
@augustas could you add instruction how can we test this task (not the unit tests only), according to our github workflow: https://oat-sa.atlassian.net/wiki/spaces/DEV/pages/11338720/Development+Workflow+with+Github

Thank you in advance :)

augustas:
> @augustas could you add instruction how can we test this task (not the unit tests only), according to our github workflow: https://oat-sa.atlassian.net/wiki/spaces/DEV/pages/11338720/Development+Workflow+with+Github
> 
> Thank you in advance :)

@zagovorichev The frontend changes that would enable the testing has not been done yet: https://oat-sa.atlassian.net/browse/TCA-395
In order to test it manually, you could enable the extra time functionality (https://github.com/oat-sa/extension-tao-proctoring/blob/master/model/implementation/DeliveryExecutionStateService.php#L58) and manually change the endpoint address here: https://github.com/oat-sa/extension-tao-proctoring/blob/master/views/js/controller/Delivery/monitoring.js#L92 to point to the new endpoint `adjustTime`. Then adding extra time would register an increase, yet without debugging you won't see the change, since the stored timer adjustments do not change the timer behavior yet.

model/execution/DeliveryExecutionList.php
```diff
@@ -102,6 +102,9 @@ private function createExecution($cachedData, $extraFields)
             'allowExtraTime' => isset($cachedData[DeliveryMonitoringService::ALLOW_EXTRA_TIME])
                 ? (bool)$cachedData[DeliveryMonitoringService::ALLOW_EXTRA_TIME]
                 : null,
+            'allowTimerAdjustment' => $this->getDeliveryExecutionManagerService()->isTimerAdjustmentAllowed(
```
SashaTAO, line 105: With this change many tests in `DeliveryExecutionListTest` fail. We need to mock `DeliveryExecutionManagerService` service in ServiceLocator mock there.

augustas: I've fixed the `DeliveryExecutionListTest` unit test.

model/execution/DeliveryExecutionManagerService.php
```diff
@@ -238,4 +235,60 @@ public function setExtraTime($deliveryExecutions, $extraTime = 0, $extendedTime
...
+            if (is_string($deliveryExecution)) {
+                $deliveryExecution = $this->getDeliveryExecutionById($deliveryExecution);
+            }
+
+            $success = false;
+            if ($this->isTimerAdjustmentAllowed($deliveryExecution)) {
+                if ($seconds > 0) {
+                    $success = $timerAdjustmentService->increase($deliveryExecution, $seconds);
+                } else {
+                    $success = $timerAdjustmentService->decrease($deliveryExecution, abs($seconds));
```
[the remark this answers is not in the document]
augustas: It's a problem with Github caching, this was updated on friday

PR #793, opened 2020-04-07
Feature/actp 342/move column to primary table

SashaTAO:
JIRA: https://oat-sa.atlassian.net/browse/ACTP-342

The goal of this development is to improve performance on proctoring screen. For that reason we move some fields common for all instances with proctoring to main `delivery_monitoring` table. With this change it won't be necessary to JOIN `kv_delievry_monitoring` table to get sessions for proctor screen. 
Add `GROUP BY` statement in query only when we join `kv_delivery_monitoring` to improve performance.

**How to test:**
- Fresh install - `test_taker_first_name, test_taker_last_name, remaining_time, extended_time, extra_time, consumed_extra_time, last_test_taker_activity` columns must should be present in delivery_monitoring table. 
- Update - `test_taker_first_name, test_taker_last_name, remaining_time, extended_time, extra_time, consumed_extra_time, last_test_taker_activity` columns created in `delivery_monitoring` table and all existing records for these keys migrated to `delivery_monitoring`. Config `config/taoProctoring/DeliveryMonitoring.conf.php` is updated with new primary columns.

In both cases `kv_delivery_monitoring` is not used to get data for proctor screen.


model/monitorCache/implementation/MonitoringStorage.php
```diff
@@ -285,12 +286,15 @@ public function find(array $criteria = [], array $options = [], $together = fals
...
-        $selectClause = "SELECT " . implode(',', $this->selectColumns);
+        $selectClause = "SELECT " . implode(','.PHP_EOL, $this->selectColumns).PHP_EOL;
 
         $sql = $selectClause . ' ' . $fromClause . PHP_EOL .
             implode(PHP_EOL, $this->joins) . PHP_EOL .
-            $whereClause . PHP_EOL .
-            'GROUP BY ' . implode(',', $this->groupColumns) . PHP_EOL;
+            $whereClause . PHP_EOL;
+
+        if ($together === true) {
```
hutnikau, line 295:
```suggestion
        if (!empty($this->groupColumns)) {
```

scripts/update/Updater.php
```diff
@@ -912,5 +912,11 @@ public function update($initialVersion)
         }
 
         $this->skip('17.3.0', '19.3.3');
+
+        if ($this->isVersion('19.3.3')) {
+            $script = 'sudo -u www-data php index.php \'oat\taoProctoring\scripts\tools\KvMonitoringMigration\' -f test_taker_first_name,test_taker_last_name,remaining_time,extended_time,extra_time,consumed_extra_time,last_test_taker_activity -d 1 -s 0 -pc';
```
augustas, line 917: When the data migration script is executed, if the columns are not present in `delivery_monitoring` table then they're created with a default type of `text`.
https://github.com/oat-sa/extension-tao-proctoring/blob/f3934acb851fe1d047fbcdf03a33801a058ac567/scripts/tools/KvMonitoringMigration.php#L206
Columns should be created before the data migration.

PR #810, opened 2020-05-14
Added `adjustment time` changes to the delivery log

zagovorichev:
Pull request summary
 
Related to : https://oat-sa.atlassian.net/browse/TCA-577
 
 - ADJUSTED_TIME added to the delivery_log
 - if request provides `reason`, it will be attached to the ADJUSTMENT_TIME data into the delivery_log
 
#### How to test
 
Send from FE part request timeAdjust with reason in the params [_example_: as it works for pause action], these data will be stored in the delivery_log table.


model/execution/DeliveryExecutionManagerService.php
```diff
@@ -285,7 +299,15 @@ public function adjustTimers(array $deliveryExecutions, $seconds)
...
                 $data->updateData([DeliveryMonitoringService::REMAINING_TIME]);
+
                 $deliveryMonitoringService->save($data);
+
+                // log adjusted time of the de
+                $adjustmentLog = [
+                    'reason' => $reason,
+                    'increment' => $seconds,
+                ];
+                $this->getDeliveryLogService()->log($deliveryExecution->getIdentifier(), 'ADJUSTED_TIME', $adjustmentLog);
```
SashaTAO, line 310: I understand that it's implemented in the same way it's done for pause/terminate requests. But in general, I think it should not happen here. I think it would be better to trigger event time adjusted and in specific listener handle this event - store log record. In this way, our code will follow the single responsibility principle and we will be able easier to handle these events asynchronously if we decide in the future.
Not critical but could be a bit better solution I think.

zagovorichev: Agree, but now I've added the log, if I do as you suggested, then I need to refactor all this class and create a lot of php unit tests, not sure that this is required by the task

model/execution/DeliveryExecutionManagerService.php
```diff
@@ -259,8 +266,14 @@ public function updateDeliveryExtendedTime(DeliveryExecutionInterface $deliveryE
...
      * @return array
+     * @throws common_Exception
+     * @throws common_exception_Error
+     * @throws common_exception_NotFound
+     * @throws common_ext_ExtensionException
+     * @throws InvalidServiceManagerException
+     * @throws QtiTestExtractionFailedException
      */
-    public function adjustTimers(array $deliveryExecutions, $seconds)
+    public function adjustTimers(array $deliveryExecutions, $seconds, array $reason = []): array
```
augustas, line 276: Phpdoc updated with unhandled exceptions but is missing the new parameter

model/execution/DeliveryExecutionManagerService.php
```diff
@@ -298,6 +320,14 @@ public function adjustTimers(array $deliveryExecutions, $seconds)
         return $result;
     }
 
+    /**
+     * @return array|object
```
augustas, line 324: For IDE search and autocompletion it would be useful to specify the interface of the object we expect to return.

model/execution/DeliveryExecutionManagerService.php
augustas: I support the suggestion by @SashaTAO, I think we were discussing this approach during planning, when discussing ticket `TCA-402`.

model/execution/DeliveryExecutionManagerService.php
```diff
@@ -285,7 +299,15 @@ public function adjustTimers(array $deliveryExecutions, $seconds)
 
                 $data = $deliveryMonitoringService->getData($deliveryExecution);
                 $data->updateData([DeliveryMonitoringService::REMAINING_TIME]);
+
                 $deliveryMonitoringService->save($data);
+
+                // log adjusted time of the de
+                $adjustmentLog = [
+                    'reason' => $reason,
+                    'increment' => $seconds,
```
augustas, line 308: Not sure if the key `increment` is good. Someone else reading the code might understand that it is only a positive value, while it could be also negative

model/execution/DeliveryExecutionManagerService.php
```diff
@@ -285,7 +299,15 @@ public function adjustTimers(array $deliveryExecutions, $seconds)
 
                 $data = $deliveryMonitoringService->getData($deliveryExecution);
                 $data->updateData([DeliveryMonitoringService::REMAINING_TIME]);
+
                 $deliveryMonitoringService->save($data);
+
+                // log adjusted time of the de
```
augustas, line 305: If you think this code block requires a comment maybe better extract it to private method.

model/execution/DeliveryExecutionManagerService.php
zagovorichev: Answered here: https://github.com/oat-sa/extension-tao-act/pull/1365#discussion_r426546078, structure was given to us by customer, for their ODS, not sure that we can change that

model/execution/DeliveryExecutionManagerService.php
zagovorichev: fixed

zagovorichev: fixed

model/execution/DeliveryExecutionManagerService.php
zagovorichev: no, I guess variable name is clear

model/execution/DeliveryExecutionManagerService.php
zagovorichev: I guess it means that any value there will be added, even if it is negative value, and formula is value + increment

model/execution/DeliveryExecutionManagerService.php
zagovorichev: fixed

model/execution/DeliveryExecutionManagerService.php
zagovorichev: done

model/deliveryLog/listener/DeliveryLogTimerAdjustedEventListener.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        $data = [
+            'reason' => $event->getReason(),
+            'increment' => $event->getSeconds(),
+            'context' => $this->getContext()
+        ];
+
+        $session = $this->getTestSessionService()->getTestSession($event->getDeliveryExecution());
+        if ($session) {
+            $data['itemId'] = $this->getCurrentItemId($event->getDeliveryExecution());
```
augustas, line 60: I agree that it might be useful to know what was the active item when time was adjusted. But currently `TEST_ADJUSTED_TIME` event details in the delivery execution history only show item id. I think it would be more useful to see category/subcategory/comment/increment and maybe item id.

model/event/DeliveryExecutionTimerAdjusted.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * @return string
+     */
+    public function getName(): string
+    {
+        return __CLASS__;
+    }
+
+    /**
+     * Returns the terminated delivery execution
```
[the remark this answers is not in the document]
zagovorichev: good catch, fixed

model/deliveryLog/listener/DeliveryLogTimerAdjustedEventListener.php
zagovorichev: I've missed this part, thank you, I've made changes according to PAUSE event (reasons only)

model/deliveryLog/listener/DeliveryLogTimerAdjustedEventListener.php
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * @param DeliveryExecutionTimerAdjusted $event
+     * @throws InvalidServiceManagerException
+     * @throws QtiTestExtractionFailedException
+     * @throws common_Exception
+     * @throws common_exception_Error
+     * @throws common_exception_NotFound
+     * @throws common_ext_ExtensionException
+     */
+    public function onTimerAdjusted(DeliveryExecutionTimerAdjusted $event): void
```
SashaTAO, line 50: I would suggest to rename this method after **what** does it do instead of **when**. We already know that it happens on timer adjusted event based on input parameter event type. If we use more informative name what does it do when this event happens it will be more clear.
As another possible improvement we could rename method to something more generic like `DeliveryStateChangeEventListeener`. Then in future we can move other delivery logging functionality (authorized, paused, terminated, etc.) here. We already have all (or almost all) required dependencies here.

model/deliveryLog/listener/DeliveryLogTimerAdjustedEventListener.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        $data = [
+            'reason' => $event->getReason(),
+            'increment' => $event->getSeconds(),
+            'context' => $this->getContext()
+        ];
+
+        $session = $this->getTestSessionService()->getTestSession($event->getDeliveryExecution());
+        if ($session) {
```
SashaTAO, line 59: We duplicate this check in two places, here and in `getCurrentItemId` method. Would be better to remove unnecessary duplication.

model/deliveryLog/event/DeliveryLogEvent.php
```diff
@@ -37,13 +37,14 @@ class DeliveryLogEvent implements Event
     public const EVENT_ID_TEST_AUTHORISE = 'TEST_AUTHORISE';
     public const EVENT_ID_TEST_TERMINATE = 'TEST_TERMINATE';
     public const EVENT_ID_TEST_IRREGULARITY = 'TEST_IRREGULARITY';
+    public const EVENT_ID_TEST_ADJUSTED_TIME = 'TEST_ADJUSTED_TIME';
```
SashaTAO, line 40: Probably would be better to use the same tense for all events.

model/deliveryLog/listener/DeliveryLogTimerAdjustedEventListener.php
zagovorichev: I think that class name should be different, it makes easier to add or replace listeners in other extensions (instead of methods rewriting)

zagovorichev: onTimerAdjusted -> adjustTime

model/deliveryLog/listener/DeliveryLogTimerAdjustedEventListener.php
zagovorichev: fixed

model/deliveryLog/event/DeliveryLogEvent.php
zagovorichev: fixed

PR #816, opened 2020-05-27
Feature/tca 579/adjustment time in de request

zagovorichev:
Pull request summary
 
Related to : https://oat-sa.atlassian.net/browse/TCA-579
 
Response from `/taoProctoring/Monitor/deliveryExecutions` contains information about adjusted time for each test session in the response
 
#### How to test
 
Check that `adjustedTime` added to the timer section of the response.

#### Dependencies
 
Requires :
 - [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1808


model/execution/DeliveryExecutionList.php
```diff
@@ -128,6 +131,42 @@ private function createExecution($cachedData, $extraFields)
         return $execution;
     }
 
+    private function getAdjustedTime(string $deliveryExecutionId): ?int
+    {
+        $deliveryExecution = $this->getDeliveryExecutionManagerService()->getDeliveryExecutionById($deliveryExecutionId);
+        $testSession = $this->getTestSessionService()->getTestSession($deliveryExecution);
+        $adjustedTime = null;
```
SashaTAO, line 138: In response to `/taoProctoring/Monitor/deliveryExecutions` we return 0 if time was not adjusted. Maybe it would be better to use here 0 as default value instead of `null`?

test/unit/model/execution/DeliveryExecutionListTest.php
```diff
@@ -251,7 +266,70 @@ public function testAdjustDeliveryExecutionsProgressStringWithCustomOption()
...
-    private function getExecutionExample()
+    public function testTimerAdjustedTimeNull(): void
+    {
+        $deliveryExecutions[] = $this->deliveryExecution;
+
+        $deliveryHelperService = new DeliveryExecutionList();
+        $deliveryHelperService->setServiceLocator($this->serviceLocatorMock);
+        /* @noinspection PhpUnhandledExceptionInspection */
+        $result = $deliveryHelperService->adjustDeliveryExecutions($deliveryExecutions);
+        $this->assertNull($result[0]['timer']['adjustedTime']);
```
SashaTAO, line 277: Assertion methods are static in phpunit 8. I think it would be better to call it as `static::assertNull()` and others the same.

model/execution/DeliveryExecutionList.php
```diff
@@ -128,6 +131,42 @@ private function createExecution($cachedData, $extraFields)
         return $execution;
     }
 
+    private function getAdjustedTime(string $deliveryExecutionId): ?int
```
SashaTAO, line 134: We have very similar method in [DeliveryExecutionManagerService.php](https://github.com/oat-sa/extension-tao-proctoring/blob/master/model/execution/DeliveryExecutionManagerService.php#L117). We also have methods to check if timer adjustment is allowed and adjust time in this service. I think it would be better to move this method to the same service as well to keep related logic in one place.

model/execution/DeliveryExecutionList.php
```diff
@@ -128,6 +131,42 @@ private function createExecution($cachedData, $extraFields)
...
+    {
+        $deliveryExecution = $this->getDeliveryExecutionManagerService()->getDeliveryExecutionById($deliveryExecutionId);
+        $testSession = $this->getTestSessionService()->getTestSession($deliveryExecution);
+        $adjustedTime = null;
+        if ($testSession) {
+            /** @var AdjustmentMap $adjustmentMap */
+            $adjustmentMap = $testSession->getTimer()->getAdjustmentMap();
```
SashaTAO, line 141: Instead of using adjustmentMap directly (imlpementation level) I would recommend to create a new method in TimerAdjustmentService and use it via this service.

model/execution/DeliveryExecutionList.php
zagovorichev: yes, sure, just wanted to note the third state when noone touched the timer

test/unit/model/execution/DeliveryExecutionListTest.php
zagovorichev: I don't see why do I need to do so, is that the best practice? it was always available with self or static, but everyone prefer to use $this in the objects and static for static... probably it makes code more flexible?

SashaTAO: `assertNull` (and all other asserts) is now a static method in phpunit. Because of that IDE always highlights it with message that "Static method should not be called dynamically"

model/execution/DeliveryExecutionList.php
zagovorichev: done

model/execution/DeliveryExecutionList.php
zagovorichev: changed

test/unit/model/execution/DeliveryExecutionListTest.php
zagovorichev: I guess that this opportunity was there, but we used self:: for the static only, so probably it should be static now, but it doesn't replace $this of phpunit

model/execution/DeliveryExecutionList.php
zagovorichev: done

test/unit/model/execution/DeliveryExecutionListTest.php
```diff
@@ -251,7 +263,18 @@ public function testAdjustDeliveryExecutionsProgressStringWithCustomOption()
...
 
-    private function getExecutionExample()
+    public function testTimerAdjustedTime(): void
+    {
+        $deliveryExecutions[] = $this->deliveryExecution;
+
+        $deliveryHelperService = new DeliveryExecutionList();
+        $deliveryHelperService->setServiceLocator($this->serviceLocatorMock);
+        /* @noinspection PhpUnhandledExceptionInspection */
+        $result = $deliveryHelperService->adjustDeliveryExecutions($deliveryExecutions);
```
augustas, line 273: On my setup phpunit fails with message:
```
Undefined index: status

/PATH/taoProctoring/model/execution/DeliveryExecutionList.php:104
```

model/execution/DeliveryExecutionManagerService.php
```diff
@@ -346,6 +360,46 @@ public function isTimerAdjustmentAllowed($deliveryExecution)
...
+     * @param string $deliveryExecutionId
+     * @return int
+     * @throws InvalidServiceManagerException
+     * @throws QtiTestExtractionFailedException
+     * @throws common_Exception
+     * @throws common_exception_Error
+     * @throws common_exception_NotFound
+     * @throws common_ext_ExtensionException
+     */
+    public function getAdjustedTime(string $deliveryExecutionId): int
```
SashaTAO, line 374: This method will return all timer adjustments, including time added as 'extendedTime' with `extended_time` LTI parameter. According to AC in ticket it must return time added by proctor separately.

zagovorichev:
> According to AC it must return time added by proctor separately from time added as `extendedTime` on LTI delivery launch.

yes, and we decided to wait your work done, then change this part

PR #820, opened 2020-05-29
Added last pause action reasons

zagovorichev:
Pull request summary
 
Related to : https://oat-sa.atlassian.net/browse/TCA-600
 
On the deliveryExecution request added last pause reasons info
 
#### How to test
 
Check `deliveryExecutions` request on the proctoring page, and find that there is a lastPauseReason data on the delivery executions which were paused by proctor

Companion PR :
 - [ ] https://github.com/oat-sa/extension-tao-proctoring/pull/821


model/execution/DeliveryExecutionList.php
```diff
@@ -125,9 +129,43 @@ private function createExecution($cachedData, $extraFields)
...
     }
 
+    private function getLastPauseReason(string $deliveryExecutionId): ?array
+    {
+        $reason = null;
+        $lastPause = $this->getDeliveryLogService()->search([
+            DeliveryLog::DELIVERY_EXECUTION_ID => $deliveryExecutionId,
+            DeliveryLog::EVENT_ID => DeliveryLogEvent::EVENT_ID_TEST_PAUSE,
+        ], [
+            'order' => 'created_at',
```
SashaTAO, line 149: I think we have auto increment ID in this table. Check indexes please if it wouldn't be better to order by ID.

zagovorichev: Not agree that usage of the identifiers is better than usage of dates, when I want to order by date. I've checked and we have an index on the created_at field

zagovorichev:
@SashaTAO @augustas Updated

test/unit/model/execution/DeliveryExecutionListTest.php
```diff
@@ -251,6 +258,29 @@ public function testAdjustDeliveryExecutionsProgressStringWithCustomOption()
         $this->assertSame('finished', $result[0]['state']['progress']);
     }
 
+    public function testLatPauseReason(): void
```
SashaTAO, line 261: Typo - `testLastPauseReason`

zagovorichev: Fixed

PR #821, opened 2020-05-29
GUI of the predefined reasons

zagovorichev:
Pull request summary
 
Related to : https://oat-sa.atlassian.net/browse/TCA-606
 
Use predefined reasons when it is possible for timeHandling form

#### How to test
 
On Change Time action in the Proctoring screen will be used reason from the last Pause action.
 
Companion PR :
 - [ ] https://github.com/oat-sa/extension-tao-proctoring/pull/820


zagovorichev:
@atsymuk it is a great catch, thank you! totally missed this case
Fixed and could you check it again, please?

PR #825, opened 2020-06-04
Fix/tca 607/catch exception from external library

SashaTAO:
JIRA: https://oat-sa.atlassian.net/browse/TCA-607

Small refactoring to make it possible to extend/modify functinality implemented in https://github.com/oat-sa/extension-tao-proctoring/pull/822

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -54,7 +54,7 @@ return array(
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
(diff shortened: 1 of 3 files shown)


SashaTAO:
> PHPUnit test does not work
> TypeError : Return value of oat\taoProctoring\model\execution\DeliveryExecutionManagerService::getSmallestMaxTimeConstraint() must be an instance of oat\taoQtiTest\models\runner\time\QtiTimeConstraint, null returned

Fixed

PR #827, opened 2020-06-11
Fix/tca 647/remaining time mismatch

augustas:
Moving several columns from `kv_delivery_monitoring` to `delivery_monitoring` table.
 
Related to : https://oat-sa.atlassian.net/browse/TCA-647
  
#### How to test

No changes in the UI or in behavior in tao community edition.

### DIFF
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -951,5 +951,12 @@ class Updater extends common_ext_ExtensionUpdater
         }
 
         $this->skip('19.10.0', '19.13.1');
+
+        if ($this->isVersion('19.13.1')) {
+            $script = 'sudo -u www-data php index.php \'oat\taoProctoring\scripts\tools\KvMonitoringMigration\' -f item_duration,stored_item_duration -d 1 -s 0 -pc -l 32';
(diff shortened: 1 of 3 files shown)


scripts/update/Updater.php
```diff
@@ -951,5 +951,12 @@ public function update($initialVersion)
         }
 
         $this->skip('19.10.0', '19.14.0');
+
+        if ($this->isVersion('19.14.0')) {
+            $script = 'sudo -u www-data php index.php \'oat\taoProctoring\scripts\tools\KvMonitoringMigration\' -f item_duration,stored_item_duration -d 1 -s 0 -pc -l 32';
+            $this->addReport(\common_report_Report::createInfo("Run script :'" . $script . "' to finish updating. Time to completion depends on table size."));
```
hutnikau, line 957: Please add a not that after launch of that script config files will be automatically changed and must be synchronized across all the instances.

augustas: During execution of `KvMonitoringMigration` script there is a message about config synchronization: `Config persisted (ONLY ON THE CURRENT SERVER. UPDATE CONFIGS ON ALL OTHER SERVERS)`.

PR #830, opened 2020-06-15
Feature/NCC-222/implement data polling for proctoring page

andrey-ponamarev:
**Related to task:**  https://oat-sa.atlassian.net/browse/NCC-222
**Description:** The polling data process for proctoring page is activated by setting time (ms) for `autoRefresh` field in the `config/taoProctoring/GuiSettings.conf.php`.
The data in table will be refreshed only if data are different with data in a previous response. We check delivery execution id and state status

**How to test:** 
- Set in `config/taoProctoring/GuiSettings.conf.php` interval for polling
```
'autoRefresh' => 3000,
```
- Login with proctor rights and navigate to the proctoring page
- Open new session and login as test taker
- As TT change the QTI test state by starting/pause QTI test. The data on the proctoring page should be reloaded in given poll interval

**Unit tests cli:** npx grunt testall

### DIFF
--- a/views/js/controller/Delivery/monitoring.js
+++ b/views/js/controller/Delivery/monitoring.js
@@ -819,7 +819,7 @@ define([
                                 title: __('Refresh the page'),
                                 label: __('Refresh'),
                                 action() {
-                                    $list.datatable('refresh', {});
+                                    $list.datatable('refresh');
                                 }
                             });
                         }
--- a/views/js/controller/Delivery/monitoring.js
+++ b/views/js/controller/Delivery/monitoring.js
@@ -742,14 +742,8 @@ define([
                         return _.isEqual(response, savedResponse);
                     }
 
-                    function startPollingData(data, ajaxConfig) {
-                        const isPollingAvailable = data.autoRefresh && ajaxConfig;
-
+                    function startPollingData(pollingInterval, ajaxConfig) {
                         stopPollingData();
-
-                        if (!isPollingAvailable) {
-                            return;
-                        }
 
                         dataPolling.setAction(function (process) {
                             const action = process.async();
@@ -769,7 +763,7 @@ define([
                                 .fail(action.resolve);
                         });
 
-                        dataPolling.setInterval(data.autoRefresh);
+                        dataPolling.setInterval(pollingInterval);
                         dataPolling.start();
                     }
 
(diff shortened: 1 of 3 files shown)


andrey-ponamarev:
Thank @jsconan for suggestions. 

> This PR was very long to review despite the change is not so complex. This was due to unrelated code style changes. If code style rework is needed, and/or cannot be avoided, it would be better to isolate it in a separate commit. Eventually, you could also mention the actual changed part.
> 
I can't understand how isolation of fixing of code style in separate commit can help you to review code 🤔  but next time I will do it and I will emphasize changes as well. I think it's better to fix code style as soon as possible rather than skip this step.
> There are a few typos in variable names, I'd suggest you activating a spellchecker in your IDE, this will prevent such trivial things.
> 
Once again thanks for pointing me to the spellchecker plugin 👍 
> The feature works well, but unfortunately, it is introducing a regression as the timer is not refreshed anymore on the proctor page. If this is something expected by your customer, please add an option.

I checked the timer on develop branch unfortunately is not working there as well  and it's not related to the current changes since the action is calling properly 🤷‍♂️

andrey-ponamarev:
@jsconan on the attached screenshot the "autoRefresh" property is calling refres of the whole table. It was exactly what we wanted to avoid, by often refreshing the proctor can't perform any action normally. Please try to test it by setting "autoRefresh" to 0.

To be on the same page here [is the code for updating the timer](https://github.com/oat-sa/extension-tao-proctoring/blob/feature/NCC-222/implement-data-polling-for-proctoring-page/views/js/controller/Delivery/monitoring.js#L168). The problem is given selector is not presenting on the page.

jsconan:
@andrey-ponamarev Thanks for the explanation. I understand you are trying to fix a usability issue, which is good, but it should not introduce a regression.
Indeed, there is a fake update of the timer, to simulate the countdown between refreshes. However, this is not the source of truth, only a helper. Any refresh **must** properly update the context info, then the fake countdown can take the relay, but a refresh is required to re-align the value.

BTW, when setting `autoRefresh` to `0`, there is a regression in your PR: clicking on the refresh button empties the table. To restore the content I need to either reload the page or update the date filter. See on this screencast: https://recordit.co/bh6WeHuGLf
image

On develop, the issue is not present, here is what I have: https://recordit.co/SDsmL2frgq
image

views/js/controller/Delivery/monitoring.js
```diff
@@ -176,1115 +178,1248 @@ define([
...
+                         */
+                        if (data.refreshBtn) {
+                            // tool: page refresh
+                            tools.push({
+                                id: 'refresh',
+                                icon: 'reset',
+                                title: __('Refresh the page'),
+                                label: __('Refresh'),
+                                action() {
+                                    $list.datatable('refresh', {});
```
jsconan, line 822: The `{}` param prevents the datatable to query the server, so that the grid will be emptied.

andrey-ponamarev: Thank you. That were for exploring, forgot to clean it 😬

alexeyvy:
Tested with Nccer

**Review Checklist**
- [x] Feature is working correctly on my local machine (if applicable)
- [x] Acceptance criteria are respected
- [x] Pull request title and description are meaningful

PR #836, opened 2020-06-24
Fix/tca 600/avoid system pause reason prepopulation

zagovorichev:
Pull request summary
 
Related to : https://oat-sa.atlassian.net/browse/TCA-600
 
The pre-population shall only occur if it c"n be "etermined t"at the"proctor almost certainly clicked on “Pause” (i.e. the “paused” session of the TT has been initiated by the Proctor and NOT by the TT losing focus)

How to test:

1. In the DE with secure mode press ESC and there will be an pause action, check that proctor doesn't see reasons when clicks on the adjustment time button
2. When proctor pauses DE by himself - then pre-populated data should be filled for the adjustment time action

### DIFF
--- a/model/execution/DeliveryExecutionList.php
+++ b/model/execution/DeliveryExecutionList.php
@@ -159,6 +159,15 @@ class DeliveryExecutionList extends ConfigurableService
         return $execution;
     }
 
+    private function isPausedByProctor($lastPause): bool
+    {
+        return isset(
+            $lastPause[0][DeliveryLog::DATA]['reason'],
+            $lastPause[0][DeliveryLog::DATA]['context']
+        )
+        && $lastPause[0][DeliveryLog::DATA]['context'] === '/taoProctoring/Monitor/pauseExecutions';
+    }
+
     private function getLastProctorPauseReason(string $deliveryExecutionId): ?array
     {
         $reason = null;
@@ -171,13 +180,7 @@ class DeliveryExecutionList extends ConfigurableService
             'limit' => 1,
         ]);
 
-        if (
-            isset(
-                $lastPause[0][DeliveryLog::DATA]['reason'],
-                $lastPause[0][DeliveryLog::DATA]['context']
-            )
-            && $lastPause[0][DeliveryLog::DATA]['context'] === '/taoProctoring/Monitor/pauseExecutions'
-        ) {
+        if ($this->isPausedByProctor($lastPause)) {
             $reason = $lastPause[0][DeliveryLog::DATA]['reason'];
         }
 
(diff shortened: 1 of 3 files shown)


model/execution/DeliveryExecutionList.php
```diff
@@ -161,7 +159,16 @@ private function createExecution($cachedData, $extraFields): array
...
     }
 
-    private function getLastPauseReason(string $deliveryExecutionId): ?array
+    private function isPausedByProctor($lastPause): bool
+    {
+        return isset(
+            $lastPause[0][DeliveryLog::DATA]['reason'],
+            $lastPause[0][DeliveryLog::DATA]['context']
+        )
+        && $lastPause[0][DeliveryLog::DATA]['context'] === '/taoProctoring/Monitor/pauseExecutions';
```
SashaTAO, line 168: It works like this. But I think it would be better to use `created_by` column from delivery_log to check if last pause event was done by Proctor. It looks more obvious to use user  role instead of controller which was called.

zagovorichev: Sorry @SashaTAO , I can't agree with you here. Now we can't divide users who launches Proctor session from the users who launches delivery execution in the test runner. So it will be not so obvious why do I see or not see predefined reasons. For example : if I use the same user to run test and to see proctors section: I will see everything predefined, even systems messages that `System generated pause`. And to other users and for me in a week it will be confusing again to understand what is not correct here, why do I see such messages.
I am categorically against the use of roles, to display the reasons, because I'm see here issues: for QA, for customers, for developers who forget or didn't know that behavior (so it is for everyone).

I already suggested you other solutions:
1. More clear as for me: I can add into the reasons new parameter (triggered on the proctor screen or something that means that), then I can use only these messages for the lastPause pre-populated reasons.
2. We can try to fix roles detection (it will be really complex solution, but we can determine that the user who has role student (test taker) - from the test runner should be marked as test taker, and if this same user makes actions as a proctor, then we can mark these actions as a made by proctor). But again, it is complex solution that takes a lot of time, and I don't think that this feature cost this time.

test/unit/model/execution/DeliveryExecutionListTest.php
```diff
@@ -292,6 +292,7 @@ public function testLastPauseReason(): void
             [
                 DeliveryLog::DATA => [
                     'reason' => ['reason'],
+                    'context' => '/taoProctoring/Monitor/pauseExecutions'
```
augustas, line 295: Maybe it would be better to use `tao_helpers_Uri::uri('pauseExecutions', 'Monitor', 'taoProctoring')` or `_url('pauseExecutions', 'Monitor', 'taoProctoring')` instead?

zagovorichev: added here: https://github.com/oat-sa/extension-tao-proctoring/pull/836/files#diff-8bee313ee340421126162b2d0d1f6a2cR165

PR #861, opened 2020-11-25
Fix/pr 89/hide extra time button when unrequired

adriloma:
**Related to:**
https://oat-sa.atlassian.net/browse/PR-90
https://oat-sa.atlassian.net/browse/PR-89
https://oat-sa.atlassian.net/browse/PR-88

**Description:**
The three issues were related with permissions configuration for the _Completed_ and _Terminated_ statuses. Also for PR-89 I included the possibility of managing statuses permissions with functions and not just boolean values, so I had to make a small refactor of the `canDo` function, which now receives the full delivery and not only the state, so different properties of it can be used to check permissions. Regarding PR-90 there was a bug that was causing to request to the server to update the timer for both, the allowed and the denied resources, when multiple sessions were selected.

**Environment:** http://test-bosa.playground.kitchen.it.taocloud.org:49151/

**How to test:**

- Using LTI launch init a TTs session for a delivery with no time limits and check as a proctor that icon to add time for the session is hidden (rows 2-4 of attached image).
- Using LTI alunch init a TTs session for a delivery with time limits, check as a proctor that icon to add time for the session is enabled, terminate or complete the session, and check that the icon to add time for the session is hidden (rows 1 and 5 of attached image).
- Select multiple sessions where adding time is enabled only for some of them, click on the icon to add time as a bulk action. (check video).

image

proctoring

### DIFF
--- a/views/js/controller/Delivery/monitoring.js
+++ b/views/js/controller/Delivery/monitoring.js
@@ -716,6 +716,18 @@ define([
                     serviceParams.context = context;
                 }
 
+                /**
+                 * Checks if button to add extra time must be available or not for a given ttSession checking
+                 * session configuration and status
+                 * @param {Object} ttSession - proctoring tts session
+                 */
+                function getAllowExtraTime(ttSession) {
+                    const sessionAllowsExtraTime = _.isNull(ttSession.allowExtraTime) || ttSession.allowExtraTime;
+                    const sessionStatusAllowsExtraTime = (ttSession.status !== 'Completed' && ttSession.status !== 'Terminated' && ttSession.timer.remaining_time !== 0);
+                    
+                    return sessionAllowsExtraTime && sessionStatusAllowsExtraTime;
+                }
+
                 return proxyExecutions.read(serviceParams).then(function(data) {
                     dataset = data.set;
                     extraFields = data.extrafields;
@@ -1098,7 +1110,7 @@ define([
                                 icon : 'time',
                                 action : timeHandling,
                                 hidden() {
-                                    var allowExtraTime = _.isNull(this.allowExtraTime) || this.allowExtraTime;
+                                    const allowExtraTime = getAllowExtraTime(this);
                                     return !canDo('time', this.state) || !allowExtraTime;
                                 }
                             }]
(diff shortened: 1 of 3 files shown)


views/js/controller/Delivery/monitoring.js
```diff
@@ -472,16 +474,17 @@ define([
...
                  */
-                function canDo(what, state) {
-                    var status;
-                    if (state && state.status) {
-                        status = _status.getStatusByCode(state.status);
-                        return status && status.can[what] === true;
+                function canDo(what, delivery) {
+                    if (delivery && delivery.state.status) {
+                        const status = _status.getStatusByCode(delivery.state.status);
+                        const canDo = _.isFunction(status.can[what]) ?status.can[what](delivery) : status.can[what];
```
andrey-ponamarev, line 485: It would be great if you could apply prettier and rename this variable to make this part of code a cleaner.
I haven't checked if every action (callback) returns `true` in case of execution `status.can[what](delivery)`, so I have some doubts about this solution 🤔

adriloma: Hi @andrey-ponamarev , no, the previous version of this function and the current version does the same: check if a given action is available for a given state. What has changed is that until now available actions for each state was a fixed truly/falsy value, and now it can be a dynamic value managed by a function. To do so I had to make two changes on this function:

1- Instead of just receiving the delivery state as a param, now it receives the full delivery so it can be used if required to evaluate a permission (state is still accesible via `delivery.state`).
2- Need to check if the permission is managed by a function so instead of returning the function I execute it (line 485).

views/js/helper/status.js
```diff
@@ -37,7 +37,9 @@ define(['lodash', 'i18n'], function(_, __){
                 report : true,
                 print : __('not finished'),
                 reactivate : __('not terminated'),
-                time : true,
+                time : function(delivery) {
```
adriloma, line 40: @andrey-ponamarev Here is an example (and the reason of the changes) of how can be permissions be dynamically managed now with a function.

jsconan (approved):
Looks good.

andrey-ponamarev (approved):

PR #865, opened 2020-12-22
Fix/PR-142/filter date with one day selected

andrey-ponamarev:
**Related to task:**  https://oat-sa.atlassian.net/browse/PR-142
**Description:** The monitoring date filter must send to server start and end day to get the proper results.
When one day is selected the end day is not sending.
**How to test:** Steps to help reviewers to test the behavior
**Steps to reproduce:** 
- Login as NCCER admin
- Enter to the child organization
- Open the proctoring page
- On page load the current day is selected and results in table is showing for the current day.
- Click on "Reload" button

**Expected behavior:** The results on the table must be the same as on page load
**Actual behavior:** The results is displaying as date filter were not applied

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -54,7 +54,7 @@ return array(
     'label' => 'Proctoring',
     'description' => 'Proctoring for deliveries',
     'license' => 'GPL-2.0',
-    'version' => '19.20.2',
+    'version' => '19.20.3',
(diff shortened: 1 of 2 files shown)


andrey-ponamarev:
The proper solution is https://github.com/oat-sa/extension-tao-proctoring/pull/866

PR #883, opened 2021-03-01
fix: Fixed QTI duration class name.

julien-sebire:
Fixed QTI duration class name in tests and phpdoc.
No functionality change.


gitromba (approved):
LGTM. Thanks @julien-sebire.

siwane (approved):

PR #917, opened 2021-12-23
Remove the test-taker "paused" dialog

yohanlaborda:
# [TR-2500](https://oat-sa.atlassian.net/browse/TR-2500)

## Changelog
- feat: add the feat flag to control the pause message
- test: add new tests for TestRunnerMessageService
- docs: add the new configuration options to README.md

## How to test
1. Set [taoProctoring](https://github.com/oat-sa/extension-tao-proctoring) up
2. Run a proctored assessment as a test taker
3. Authorize the assessment as a proctor
4. Pause the assessment as the test taker
5. Observe that the message is sent to the test taker
6. Set `FEATURE_FLAG_PROCTOR_NOT_LAUNCH_PAUSE_MESSAGE="true"` environment variable
8. Restart the assessment as the test taker
9. Pause the assessment as the test taker
10. Observe that it does not show the message to the test-taker and reload the page

### DIFF
--- a/model/implementation/TestRunnerMessageService.php
+++ b/model/implementation/TestRunnerMessageService.php
@@ -95,7 +95,7 @@ class TestRunnerMessageService extends QtiRunnerMessageService
      */
     protected function getPausedStateMessage(AssessmentTestSession $testSession)
     {
-        if ($this->cantReturnPauseStateMessage()) {
+        if (false === $this->isPausedDialogRequired()) {
             return '';
         }
 
@@ -120,9 +120,9 @@ class TestRunnerMessageService extends QtiRunnerMessageService
         return parent::getTerminatedStateMessage($testSession);
     }
 
-    private function cantReturnPauseStateMessage(): bool
+    private function isPausedDialogRequired(): bool
     {
-        return $this->getFeatureFlagChecker()->isEnabled(self::FEATURE_FLAG_PROCTOR_NOT_LAUNCH_PAUSE_MESSAGE);
+        return false === $this->getFeatureFlagChecker()->isEnabled(self::FEATURE_FLAG_SKIP_PAUSED_ASSESSMENT_DIALOG);
     }
 
     private function getFeatureFlagChecker(): FeatureFlagCheckerInterface
--- a/model/implementation/TestRunnerMessageService.php
+++ b/model/implementation/TestRunnerMessageService.php
@@ -23,7 +23,8 @@
 
 namespace oat\taoProctoring\model\implementation;
 
-use oat\taoProctoring\model\ProctorService;
+use oat\tao\model\featureFlag\FeatureFlagChecker;
+use oat\tao\model\featureFlag\FeatureFlagCheckerInterface;
 use oat\taoQtiTest\models\runner\QtiRunnerMessageService;
 use qtism\runtime\tests\AssessmentTestSession;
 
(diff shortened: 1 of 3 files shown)


test/unit/model/implementation/TestRunnerMessageServiceTest.php
```diff
@@ -0,0 +1,132 @@
...
+            ]
+        ];
+    }
+
+    /**
+     * @param array $options
+     * @param ServiceManager $serviceManager
+     * @return TestRunnerMessageService
+     */
+    private function getClassTestRunnerMessageService($serviceManager, $options = []): TestRunnerMessageService
```
poyuki, line 100: probably we can add type hints to arguments and remove phpdock block, WDYT?

yohanlaborda: Good idea

model/implementation/TestRunnerMessageService.php
```diff
@@ -36,6 +37,11 @@
  */
 class TestRunnerMessageService extends QtiRunnerMessageService
 {
+    /**
+     * @var string Controls whether a pause message is not sent on status, default false
+     */
+    private const FEATURE_FLAG_PROCTOR_NOT_LAUNCH_PAUSE_MESSAGE = 'FEATURE_FLAG_PROCTOR_NOT_LAUNCH_PAUSE_MESSAGE';
```
wazelin, line 43: Let's rename this to `FEATURE_FLAG_SKIP_PAUSED_ASSESSMENT_DIALOG`

model/implementation/TestRunnerMessageService.php
```diff
@@ -110,4 +120,13 @@ protected function getTerminatedStateMessage(AssessmentTestSession $testSession)
         return parent::getTerminatedStateMessage($testSession);
     }
 
+    private function cantReturnPauseStateMessage(): bool
+    {
+        return $this->getFeatureFlagChecker()->isEnabled(self::FEATURE_FLAG_PROCTOR_NOT_LAUNCH_PAUSE_MESSAGE);
```
wazelin, line 125: Please don't use negations in the methods or variable names, it creates a lot of confusion.
Let's reverse this, and name it `isPausedDialogRequired()`. WDYT?
```suggestion
    private function isPausedDialogRequired(): bool
    {
        return !$this->getFeatureFlagChecker()->isEnabled(self::FEATURE_FLAG_PROCTOR_NOT_LAUNCH_PAUSE_MESSAGE);
```

Obviously, the users of the method have to take that into account, and reverse their conditions.

wazelin (changes requested):

wazelin:
In my case, I can see an error message instead of being redirected to the `paused` screen, when using this feature branch.
I'm using the latest versions of the other extensions.

https://user-images.githubusercontent.com/2943256/147344428-9bef577b-6150-40dc-a86a-bf2aad0a78dc.mov

model/implementation/TestRunnerMessageService.php
yohanlaborda: Good

augustas:
I believe this PR can be closed, it's better to rely on current way of test runner feature flag configuration, as it was done here: https://github.com/oat-sa/extension-tao-testqti/pull/2158

